PR #8, opened 2020-08-25
WIFI-329: enabled resource limits for cassandra

eugenetaranov-opsfleet:


rahulsharma2020 (approved):

rahulsharma2020:
We need to keep an eye on Cassandra pod with these resource-limits. 
In my opinion, for a small system, this should be ok.

eugenetaranov-opsfleet:
> We need to keep an eye on Cassandra pod with these resource-limits.
> In my opinion, for a small system, this should be ok.

Definitely! The reason for adding limits - k8s is unable to properly schedule it when nodes are relatively small or there are other resource intensive workloads - cassandra gets killed by OOM.

PR #9, opened 2020-08-28
cloudsdk for minikube local environment

eugenetaranov-opsfleet:


tip-wlan/charts/opensync-gw-cloud/templates/deployment.yaml
```diff
@@ -28,139 +28,145 @@ spec:
...
-              exit 1
-            fi        
-      {{- end }}
+      - name: wait-for-services
+        image: opsfleet/depends-on
+        args:
+        - "-service=tip-wlan-opensync-mqtt-broker" # fixme
+        - "-service=tip-wlan-wlan-prov-service" # fixme
+        - -check_interval=5
+      #   - name: {{ include "common.name" . }}-mqtt-readiness
```
leonidlm: @eugenetaranov if you're sure the commented out code isn't needed anymore, let's delete it entirely instead of commenting out. We have git history if we would like to revert back or check what was deleted in the future.

leonidlm (approved):
@eugenetaranov looks good, but I suggest you also get @rahulsharma2020 feedback to make sure we aren't introducing breaking changes.

tip-wlan/charts/opensync-gw-cloud/templates/deployment.yaml
rahulsharma2020: @eugenetaranov : I am not sure what's the way to query a service readiness in the `opsfleet/depends-on` image but we need to make sure that prov, ssc and mqtt services are running and responding to requests (/ping for prov and ssc and 'Aliveness check' for Mqtt) before we start Opensync-Gw svc.
Also when running in Integrated env (prov, ssc and spc as one process), we need to make sure that the IntegratedComponent responds to /ping .

tip-wlan/charts/wlan-prov-service/resources/config/datasource.properties
```diff
@@ -1,11 +1,11 @@
 singleDataSource.url=jdbc:postgresql://tip-wlan-postgresql:5432/prov_db
-singleDataSource.username=DUMMY_POSTGRES_TIP_USER
-singleDataSource.password=DUMMY_POSTGRES_TIP_PASSWORD
+singleDataSource.username=tip_user
+singleDataSource.password=tip_password
```
rahulsharma2020: Do you plan to commit the passwords in the charts?

A better approach would be to convert such files into configmaps, get the values of these parameters from values.yaml and then override these from dev-*.yaml file.

tip-wlan/charts/opensync-gw-cloud/templates/deployment.yaml
eugenetaranov-opsfleet: > what's the way to query a service readiness in the opsfleet/depends-on image

it does not support this, it relies on pod passing k8s healthchecks

> we need to make sure that prov, ssc and mqtt services are running and responding to requests 
this is a task for startup/liveness/readiness probes to make sure the service is alive

having liveness probes defined in 2 places does not improve application health but makes it harder to maintain and slowdowns startup

what do you think @rahulsharma2020 , @leonidlm , @cbuschfb ?

rahulsharma2020: @eugenetaranov-opsfleet: I would need an example to understand how you would be passing the K8S healthchecks.
Also, the liveliness defined at 2 places - don't understand that comment. Can you give pointers to what you are referring to?

tip-wlan/charts/wlan-prov-service/resources/config/datasource.properties
eugenetaranov-opsfleet: they are already hardcoded everywhere .. refactoring this may make another good PR @rahulsharma2020

tip-wlan/charts/opensync-gw-cloud/templates/deployment.yaml
leonidlm: @rahulsharma2020 I see that the services you've mentioned have the proper readiness probs defined. For example, the prov service has its readiness probe run the `/ping` HTTP request you've mentioned (https://github.com/Telecominfraproject/wlan-cloud-helm/blob/master/tip-wlan/charts/wlan-prov-service/templates/deployment.yaml#L149).

The above means that we don't need to run another explicit `/ping` check on opensync-gw's startup, and instead just rely on `depends-on` to wait until prov service will have at least one endpoint registered (== at least one pod moved to the ready state based on it's readiness prob's status).

Do you see another reason why relying on another service's readiness checks won't be enough? (assuming the readiness probes were defined correctly).

rahulsharma2020: I see. Thanks for the explanation @leonidlm, understand it now.
In that case, the /ping for `prov` and `ssc` processes are good. 
However for mqtt, it's different:
- the readiness probe currently checks for `tcpSocket` on a particular port. 
- Whereas for checking mqtt readiness from Opensync-gw pod, we use the mosquitto_pub (a mosquitto client - https://mosquitto.org/man/mosquitto_pub-1.html) to make sure it's alive. 
So to use the `depends-on` image for MQTT, we should change it's readiness probe to use mosquitto_pub instead.

leonidlm: @rahulsharma2020 sounds like a good idea! Using mosquitto_pub instead of tcpSocket for readiness probe is probably a good idea anyway, unrelated to depends-on. @eugenetaranov WDYT?

eugenetaranov-opsfleet: @leonidlm already added here as startupProbe https://github.com/Telecominfraproject/wlan-cloud-helm/blob/7fee3eca5d2673d7225840339394b1358c260772/tip-wlan/charts/opensync-mqtt-broker/templates/statefulset.yaml#L84-L97

tip-wlan/charts/opensync-mqtt-broker/templates/statefulset.yaml
```diff
@@ -127,6 +149,9 @@ spec:
       - name: opensync-mqtt-broker-conf
         configMap:
             name: mosquitto-config
+      - name: certificates
+        secret:
+          secretName: tip-wlan-opensync-gw-cloud-certs
```
rahulsharma2020: Don't prefix the release-name to secretName

tip-wlan/charts/opensync-gw-cloud/templates/deployment.yaml
```diff
@@ -28,139 +28,45 @@ spec:
...
-            else
-              echo ${url} service failed to respond after 50 secs
-              exit 1
-            fi        
-      {{- end }}
+      - name: wait-for-services
+        image: opsfleet/depends-on
+        args:
+        - "-service=tip-wlan-opensync-mqtt-broker" # fixme
+        - "-service=tip-wlan-wlan-prov-service" # fixme
```
rahulsharma2020: You should add ssc as well here.
Also for integrated deployment, use `wlan-integrated-cloud-component-service`.

Also remove the Release-Name (`tip-wlan`) from the services name

tip-wlan/charts/opensync-gw-cloud/templates/deployment.yaml
rahulsharma2020: Thanks @eugenetaranov-opsfleet . Looks good. Should we just put that in readiness/liveliness probe instead of startupProbe or do you see any issue there?

eugenetaranov-opsfleet: @rahulsharma2020 I think it's sufficient to have it checked only once at startup but if you feel like it makes sense to check it every minute or so let's do it. The only thing we need to make sure is that test messages auto expire and do not fill disk or cause other issues.

tip-wlan/charts/wlan-prov-service/resources/config/datasource.properties
eugenetaranov-opsfleet: @rahulsharma2020 does this look ok to you? https://github.com/Telecominfraproject/wlan-cloud-helm/pull/9/files#diff-d606bccc7c86fa7a06b288c3d53be757668c4d61b649419ee337fdfceeac7b4cR8

tip-wlan/charts/opensync-gw-cloud/templates/deployment.yaml
```diff
@@ -28,139 +28,47 @@ spec:
...
-              echo ${url} service failed to respond after 50 secs
-              exit 1
-            fi        
-      {{- end }}
+      - name: wait-for-services
+        image: opsfleet/depends-on
+        args:
+        - "-service={{ .Release.Name }}-opensync-mqtt-broker"
+        - "-service={{ .Release.Name }}-wlan-prov-service"
+        - "-service={{ .Release.Name }}-wlan-prov-service"
```
rahulsharma2020: Prov service should not be repeated.

tip-wlan/charts/opensync-gw-cloud/templates/deployment.yaml
```diff
@@ -28,139 +28,47 @@ spec:
...
+        - "-service={{ .Release.Name }}-wlan-ssc-service"
+        - -check_interval=5
+      # {{- if .Values.global.integratedDeployment }}
+      #   - name: {{ include "common.name" . }}-readiness-int-cloud
```
rahulsharma2020: Also, we may want to add dependency on integrated-cloud-component if `integratedDeployment` flag is true

tip-wlan/charts/opensync-mqtt-broker/templates/statefulset.yaml
```diff
@@ -76,13 +76,35 @@ spec:
           readinessProbe:
             tcpSocket:
               port: {{ .Values.service.port1 }}
-            initialDelaySeconds: {{ .Values.probes.readinessProbe.initialDelaySeconds }}
-            timeoutSeconds: {{ .Values.probes.readinessProbe.timeoutSeconds }}
+            initialDelaySeconds: {{ .Values.probes.livenessProbe.initialDelaySeconds }}
```
rahulsharma2020: Did we change the `.Values.probes.readinessProbe.` to `.Values.probes.livenessProbe.` under readinessProbe Or is it just my diff-viewer not showing correctly?

tip-wlan/charts/wlan-prov-service/templates/configmap.yaml
```diff
@@ -4,4 +4,16 @@ metadata:
...
+    singleDataSource.username={{ .Values.creds.postgres.singleDataSourceUsername }}
+    singleDataSource.password={{ .Values.creds.postgres.singleDataSourcePassword }}
+    singleDataSource.driverClass=org.postgresql.Driver
+    singleDataSource.ssl=true
+    singleDataSource.sslmode=verify-ca
+    singleDataSource.sslcert=/opt/tip-wlan/certs/postgresclientcert.pem
+    singleDataSource.sslfactory=org.postgresql.ssl.LibPQFactory
+    singleDataSource.sslkey=/opt/tip-wlan/certs/postgresclient.p12
+    singleDataSource.sslrootcert=/opt/tip-wlan/certs/cacert.pem
+    singleDataSource.sslkeypassword=mypassword
```
rahulsharma2020: Please change this to a variable in secret that we can read from

tip-wlan/charts/wlan-ssc-service/resources/config/cassandra-application.conf
```diff
@@ -14,14 +14,14 @@ datastax-java-driver {
         class = DefaultSslEngineFactory    
         hostname-validation = false
         truststore-path = /opt/tip-wlan/certs/truststore.jks
-        truststore-password = DUMMY_PASSWORD
+        truststore-password = mypassword
```
rahulsharma2020: Either convert this file into ConfigMap (as done for prov-service) or remove the actual password from the file

tip-wlan/resources/environments/dev-amazon-tip.yaml
```diff
@@ -116,10 +116,10 @@ wlan-spc-service:
     sslKeystorePassword: mypassword
     sslTruststorePassword: mypassword
 nginx-ingress-controller:
-  enabled: true
+  enabled: false
```
rahulsharma2020: Why is nginx disabled? We may have issues with UI

eugenetaranov-opsfleet: that came from cicd deployment, we had conflicts with nginx-ingress already installed there for monitoring stuff

tip-wlan/charts/wlan-prov-service/templates/configmap.yaml
eugenetaranov-opsfleet: please elaborate

tip-wlan/resources/environments/dev-amazon-tip.yaml
rahulsharma2020: Hmmm... In that case:
- either we need to resolve the conflict, so both can run
- or we need to configure the other nginx to forward the traffic coming for a service on our cluster appropriately.

tip-wlan/charts/wlan-prov-service/templates/configmap.yaml
rahulsharma2020: I meant create a secret something this: https://github.com/Telecominfraproject/wlan-cloud-helm/blob/master/tip-wlan/charts/wlan-prov-service/templates/secret.yaml#L23 and then read the variable's value in this configMap.
That way, you can put in DUMMY_VALUE values.yaml for this service and then override it in override file

tip-wlan/resources/environments/dev-amazon-tip.yaml
eugenetaranov-opsfleet: I think they're so many problems with these recipes and it's important to keep changes small and manageable, not trying to fix all issues at once.
I'll revert this particular file to the original state as its not related to the minikube setup.

tip-wlan/resources/environments/dev-local.yaml
```diff
@@ -0,0 +1,197 @@
...
+
+# Docker registry secret
+dockerRegistrySecret: ewoJImF1dGhzIjogewoJCSJ0aXAtdGlwLXdsYW4tY2xvdWQtZG9ja2VyLXJlcG8uamZyb2cuaW8iOiB7CgkJCSJhdXRoIjogImRHbHdMWEpsWVdRNmRHbHdMWEpsWVdRPSIKCQl9Cgl9LAoJIkh0dHBIZWFkZXJzIjogewoJCSJVc2VyLUFnZW50IjogIkRvY2tlci1DbGllbnQvMTkuMDMuOCAobGludXgpIgoJfQp9
+#################################################################
+# Enable/disable and configure helm charts (ie. applications)
+# to customize the TIP-WLAN deployment.
+#################################################################
+opensync-gw-static:
+  enabled: false
+common:
```
rahulsharma2020: Don't need this block in dev-local.yaml

tip-wlan/resources/environments/dev-local.yaml
```diff
@@ -0,0 +1,197 @@
...
+  persistence:
+    enabled: true
+    storageClass: standard
+cassandra:
+  enabled: true
+  image:
+    debug: true
+  # config:
+  #   replicaCount: 1
+  #   seedCount: 1
```
rahulsharma2020: Perhaps removing them is better

tip-wlan/resources/environments/dev-local.yaml
eugenetaranov-opsfleet: removed ^common:

tip-wlan/charts/kafka/resources/config/certs/keystore_creds
```diff
@@ -1 +1 @@
-DUMMY_PASSWORD
\ No newline at end of file
+mypassword
```
rahulsharma2020: Please revert this to DUMMY_PASSWORD, (until we change this file into a ConfigMap/Secret and read it from override file)

tip-wlan/charts/kafka/resources/config/certs/truststore_creds
```diff
@@ -1 +1 @@
-DUMMY_PASSWORD
\ No newline at end of file
+mypassword
```
rahulsharma2020: Please revert this to DUMMY_PASSWORD, (until we change this file into a ConfigMap/Secret and read it from override file)

tip-wlan/charts/wlan-prov-service/templates/configmap.yaml
```diff
@@ -4,4 +4,16 @@ metadata:
...
```
rahulsharma2020: Please read its values from values.yaml instead of hardcoding the password, now that it's in a ConfigMap. Also in the values.yaml, we should put in DUMMY_PASSWORD and overriden it from the override-file.

tip-wlan/charts/wlan-ssc-service/values.yaml
```diff
@@ -107,8 +107,8 @@ affinity: {}
 creds:
   enabled: true
   sslKeyPassword: DUMMY_PASSWORD
-  sslKeystorePassword: DUMMY_PASSWORD
-  sslTruststorePassword: DUMMY_PASSWORD
+  sslKeystorePassword: mypassword
```
rahulsharma2020: Please revert this to DUMMY_PASSWORD; these values should be overriden from override-file, not in the chart directly.

tip-wlan/charts/wlan-ssc-service/values.yaml
```diff
@@ -107,8 +107,8 @@ affinity: {}
...
+  sslTruststorePassword: mypassword
```
rahulsharma2020: Please revert this to DUMMY_PASSWORD; these values should be overriden from override-file, not in the chart directly.

tip-wlan/charts/wlan-ssc-service/templates/configmap.yaml
```diff
@@ -4,4 +4,30 @@ metadata:
...
+        hostname-validation = false
+        truststore-path = /opt/tip-wlan/certs/truststore.jks
+        truststore-password = {{ .Values.creds.sslTruststorePassword }}
+        keystore-path = /opt/tip-wlan/certs/cassandra_server_keystore.jks
+        keystore-password = {{ .Values.creds.sslKeystorePassword }}
+      }
+
+      advanced.auth-provider {
+        class = PlainTextAuthProvider
+        username = tip_user
```
rahulsharma2020: Ideally even this should come from values.yaml as DUMMY_USER and then overriden in the override file.

tip-wlan/charts/wlan-ssc-service/templates/configmap.yaml
```diff
@@ -4,4 +4,30 @@ metadata:
...
+        password = tip_password
```
rahulsharma2020: Ideally even this should come from values.yaml as DUMMY_PASSWORD and then overriden in the override file.

rahulsharma2020 (approved):

tip-wlan/charts/wlan-prov-service/templates/configmap.yaml
```diff
@@ -4,4 +4,16 @@ metadata:
...
+    singleDataSource.username={{ .Values.creds.postgres.singleDataSourceUsername }}
+    singleDataSource.password={{ .Values.creds.postgres.singleDataSourcePassword }}
+    singleDataSource.driverClass=org.postgresql.Driver
+    singleDataSource.ssl=true
+    singleDataSource.sslmode=verify-ca
+    singleDataSource.sslcert=/opt/tip-wlan/certs/postgresclientcert.pem
+    singleDataSource.sslfactory=org.postgresql.ssl.LibPQFactory
+    singleDataSource.sslkey=/opt/tip-wlan/certs/postgresclient.p12
+    singleDataSource.sslrootcert=/opt/tip-wlan/certs/cacert.pem
+    singleDataSource.sslkeypassword=DUMMY_PASSWORD
```
rahulsharma2020, line 18: Should be read from values.yaml and then overriden in an override file.

PR #11, opened 2020-10-02
Aj

AkshayJagadish-ne:
WIFI-845: Support Kubernetes Deployment Versioning


tip-wlan/charts/wlan-ssc-service/values.yaml
```diff
@@ -12,6 +12,7 @@ replicaCount: 1
 image:
   name: wlan-ssc-service
   tag: 0.0.1-SNAPSHOT
+  timestamp:
```
rahulsharma2020: As discussed, please remove the timestamp from every chart

rahulsharma2020 (approved):

PR #12, opened 2020-10-07
Changes to support Port-forwarding-gateway properties: PF_GATEWAY_EXT_HOST and  PF_GATEWAY_EXT_PORT

AkshayJagadish-ne:


tip-wlan/resources/environments/dev-amazon-tip.yaml
```diff
@@ -120,6 +120,11 @@ wlan-port-forwarding-gateway-service:
   enabled: true
   creds:
     websocketSessionTokenEncKey: MyToKeN0MyToKeN1
+  env:
+    pfGateway:
+    externallyVisibleHost: api.wlan.demo.lab.wlan.tip.build
```
rahulsharma2020: Indentation needs fixing or is it just this tool?

tip-wlan/charts/wlan-port-forwarding-gateway-service/values.yaml
```diff
@@ -170,3 +170,5 @@ env:
   pfGateway:
     service: wlan-port-forwarding-gateway-service
     port: 7070
+    externallyVisibleHost: pfg.example.org
```
rahulsharma2020: Though the properties are fine here but in general I tend to keep them out of the "env" since we are controlling values under "env" from the template defined under common/templates/env.tpl file.
I would recommend that we do something similar to what we did in opensync-gw-cloud - see https://github.com/Telecominfraproject/wlan-cloud-helm/blob/master/tip-wlan/charts/opensync-gw-cloud/values.yaml#L120-L122

tip-wlan/resources/environments/dev-amazon-tip.yaml
AkshayJagadish-ne: The indentation needs fixing. Thank you very much! I will double-check the indentations for other files too

tip-wlan/charts/wlan-port-forwarding-gateway-service/values.yaml
AkshayJagadish-ne: Thank you Rahul, I will make this change. And the Values in deployment.yaml would be accessible as .Values.pfGateway.externallyVisiblePort    instead of  .Values.env.pfGateway.externallyVisiblePort

rahulsharma2020: I would remove even pfGateway. So something as follows:
```
externallyVisible:
    host:
    port:
```
since we already have `wlan-port-forwarding-gateway-service` under which these properties are defined, so no need to repeat. What do you think?

AkshayJagadish-ne: Yes, even I think pfGateway is not necessary.  I will make these changes. Thank you very much!

PR #16, opened 2020-10-19
Feature/helmfile

4c74356b41:
this is a PR to start using Helmfile. This improves the workflow in many ways:

1. Gives better flexibility to deploy individual components using labels
2. Using latest bitnami charts for external components (and some other latest charts), so we dont have to maintain our own chart forks
3. more flexibility with variable management using different helmfile features
4. centralized secret management (WIP)

This new flow doesnt interrupt existing helm flow, but for some time while we migrate, we have to support 2 different code bases.

This is somewhat of a WIP so that we can start reviewing this PR (as it is quite big), most of the things are working, but we need to test this


leonidlm:
@4c74356b41 now I understand what you meant by "huge" PR. Let's try to see how we can split this up into workable and smaller PRs. By the first look at everything, I might have a few ideas.

4c74356b41:
will try and rework this into smaller pieces, this draft left for reference

PR #17, opened 2020-10-22
fixed missing values in wlan-prov

eugenetaranov-opsfleet:


tip-wlan/charts/wlan-prov-service/values.yaml
```diff
@@ -65,7 +65,9 @@ creds:
       password: DUMMY_POSTGRES_PASSWORD
```
rahulsharma2020, line 65: This is now duplicated. May be, if not in this PR, remove it and update charts to use the new properties added below.

rahulsharma2020 (approved):

PR #19, opened 2020-10-27
local development env with minikube

eugenetaranov-opsfleet:


4c74356b41 (approved):
lgtm

README.md
```diff
@@ -45,3 +45,129 @@ This repository contains helm charts for various deployment types of the tip wla
...
+done
+```
+
+Run minikube:
+
+```minikube start --memory=10g --cpus=4 --driver=virtualbox --extra-config=kubelet.serialize-image-pulls=false --extra-config=kubelet.image-pull-progress-deadline=3m0s --docker-opt=max-concurrent-downloads=10```
+
+Deploy CloudSDK chart:
+
+```helm install tip-wlan tip-wlan -f tip-wlan/resources/environments/dev-local.yaml -n default```
```
leonidlm: Maybe use `helm upgrade --install` here so people who don't know Helm can re-run this command multiple times.

README.md
```diff
@@ -45,3 +45,129 @@ This repository contains helm charts for various deployment types of the tip wla
...
+minikube cache reload ## reload images from the upstream
+eval $( minikube docker-env )
+for img in $( docker images --format '{{.Repository}}:{{.Tag}}' | egrep 'busybox|alpine|confluentinc/cp-kafka|zookeeper|k8s.gcr.io/pause|nginx/nginx-ingress|bitnami/cassandra|bitnami/postgresql|postgres|bitnami/minideb' ); do
+	minikube cache add $img;
+done
+```
+
+Run minikube:
+
+```minikube start --memory=10g --cpus=4 --driver=virtualbox --extra-config=kubelet.serialize-image-pulls=false --extra-config=kubelet.image-pull-progress-deadline=3m0s --docker-opt=max-concurrent-downloads=10```
```
leonidlm, line 95: How about mentioning that we can run this with `--driver=none` to make this setup even leaner. I think in some OSes like Linux this is even preferable and the startup will be much faster. You won't need to specify the memory/cpu requests too.

README.md
```diff
@@ -45,3 +45,129 @@ This repository contains helm charts for various deployment types of the tip wla
...
+
+Deploy CloudSDK chart:
+
+```helm install tip-wlan tip-wlan -f tip-wlan/resources/environments/dev-local.yaml -n default```
+
+Wait a few minutes, when all pods are in `Running` state, obtain web ui link with `minikube service tip-wlan-wlan-cloud-static-portal -n tip --url`, open in the browser. Importing or trusting certificate might be needed.
+
+Services may be exposed to the local machine and local network with ssh, kubectl or kubefwd port forwarding, needs to be repeated for each service, please examples below:
+
+Kubefwd:
```
leonidlm, line 107: I would mention what Kubefwd does and also note that this is for development purposes only and that in real use cases exposing the services "outside" Kubernetes infra will be done through the loadbalancer or with MetalLB

README.md
```diff
@@ -45,3 +45,129 @@ This repository contains helm charts for various deployment types of the tip wla
...
+
+Download latest release from https://github.com/eugenetaranov/kubefwd/releases and run the binary.
+
+Forward to all interfaces:
+
+```
+sudo kubefwd services --namespace tip -l "app.kubernetes.io/name in (nginx-ingress-controller,wlan-portal-service,opensync-gw-cloud,opensync-mqtt-broker)" --allinterfaces --extrahosts wlan-ui-graphql.wlan.local,wlan-ui.wlan.local
+```
+
+Kubectl forwarding:
```
leonidlm: Mention that this is an alternative to using Kubefwd, that's not immediately obvious.

leonidlm (approved):

README.md
eugenetaranov-opsfleet: Sure, will add a note mentioning other drivers.

> You won't need to specify the memory/cpu requests too.

In this case the system may freeze if there is not enough RAM. Might be better to fail earlier due to insufficient memory instead of having OOM storm..

tip-wlan/charts/opensync-gw-cloud/resources/config/logback.xml
```diff
@@ -73,6 +73,7 @@
 
   <root level="WARN">
     <appender-ref ref="logfile"/>
+    <appender-ref ref="stdout"/>
```
rahulsharma2020, line 76: I think this should be set to `logfile` as default since we have quite a few tweaks/added functionality (encoders, rollingPolicy etc) to make it easier to maintain. Should be set to `stdout` only for special cases (to look at the logs during startup). Why are both enabled?

tip-wlan/charts/kafka/templates/secret.yaml
```diff
@@ -1,8 +1,18 @@
...
   namespace: {{ include "common.namespace" . }}
+  labels:
+    app: {{ template "common.name" . }}
+    chart: {{ template "common.chart" . }}
+    release: {{ .Release.Name | quote }}
 type: Opaque
 data:
-{{ tpl (.Files.Glob "resources/config/certs/*").AsSecrets . | indent 2 }}
\ No newline at end of file
+  truststore_creds: {{ .Values.creds.sslTruststorePassword | b64enc }}
```
rahulsharma2020, line 13: This is good. Thank you for removing the files and instead getting the values from secrets

tip-wlan/charts/cassandra/templates/secret.yaml
```diff
@@ -24,9 +24,9 @@ metadata:
 type: Opaque
 data:
   truststore: {{ .Files.Get "resources/config/certs/truststore.jks" | b64enc }}
-  truststore-password: {{ .Files.Get "resources/config/certs/truststore_creds" | b64enc }}
+  truststore-password: {{ .Values.creds.sslTruststorePassword | b64enc }}
```
rahulsharma2020, line 27: This is good. Thank you for removing the files and instead getting the values from secrets

tip-wlan/charts/opensync-gw-static/resources/config/logback.xml
```diff
@@ -68,7 +68,7 @@
    -->
    
   <root level="WARN">
-<!--    <appender-ref ref="stdout"/>-->
+    <appender-ref ref="stdout"/>
```
rahulsharma2020, line 71: Same comment as above (and also for all the changes below). Why are both enabled?

tip-wlan/charts/kafka/values.yaml
```diff
@@ -353,3 +353,8 @@ zookeeper:
   #       labelSelector:
   #         matchLabels:
   #           release: zookeeper
+
+creds:
```
rahulsharma2020, line 357: While we are converting `cassandra-application.conf` and Postgres datasource (`datasource.properties`) to configMaps, what are your views on `admin-client.properties` file under Kafka? Should that get converted into ConfigMap as well

eugenetaranov-opsfleet: thanks! updated it too

tip-wlan/charts/opensync-gw-cloud/resources/config/logback.xml
4c74356b41: actually you are right. we should just drop the logfile appender, it doesnt make any sense in k8s environment

leonidlm: @rahulsharma2020 we usually suggest that all microservices will send logs to stdout by default. In Kubernetes, this is considered best practice because then you can use cluster-wide log shipping, such as sending all the logs to ES and Kibana, basically mimicking the suggestions of 12-factor apps (https://12factor.net/logs).
If you feel that's a much more significant change to how we handle things at the moment, we can schedule a call to discuss all the implications of this change.

eugenetaranov-opsfleet: I think there are at least 2 reasons why we need stdout:
1. simplify deployment and maintenance - logs (at least ERROR) should be available via `kubectl logs ..` commands, anyone who troubleshoots pod start failures will appreciate this.
2. local logging is an antipattern in cloud/k8s alike environments where pods come and go.. We'd need to send them to some central location, attach a sidecar to gather these logs - a lot of extra steps instead of just sending to stdout.

Do you have any reasons for the opposite? @rahulsharma2020

rahulsharma2020: Thanks @leonidlm, @eugenetaranov-opsfleet. Agree with your explanation.

My take on this: Yes, we would definitely need to look at `stdout` during **startup** - as I mentioned in my above comment and is very useful. However, once the pod is up and running, there would be lots of logs and looking at them using `kubectl logs` isn't easy - especially for services like opensync-gw which spits out tons of logs from APs management. (And these logs could easily be uncontrollable if the number of APs managed go in the range of hundreds/thousands.) With `logfile` approach, we can limit size of each, copy them on our system and look at them individually and chronologically to check for errors from buggy APs.
Until we have ES/Kibana running to take care of logging from multiple different services, we can go either way.

As originally intended in the PR, let's keep both for now and we can check and decide later if we want to remove anything.

eugenetaranov-opsfleet: thanks for explanation @rahulsharma2020 !
I am not sure if there is a way to disable logs after startup ..
what we might consider later - send ERROR logs to stdout - so they would be useful to maintain the services, and send all logs to some collector forwarding them to elasticsearch ..

rahulsharma2020: Agree @eugenetaranov-opsfleet

rahulsharma2020 (approved):

PR #21, opened 2020-11-06
WIFI-991: alb ingress

eugenetaranov-opsfleet:
Installation:

```
helm upgrade --install aws-load-balancer-controller eks/aws-load-balancer-controller -n kube-system --set clusterName=tip-wlan-testcluster --set serviceAccount.annotations."eks\.amazonaws\.com/role-arn"="arn:aws:iam::289708231103:role/tip-wlan-testcluster-alb-ingress"
```

- creates shared alb for all Ingress based services;
- creates dedicated nlb for each service;

### DIFF
--- a/tip-wlan/charts/wlan-cloud-graphql-gw/templates/ingress.yaml
+++ b/tip-wlan/charts/wlan-cloud-graphql-gw/templates/ingress.yaml
@@ -32,6 +32,12 @@ spec:
     - host: {{ .host | quote }}
       http:
         paths:
+        {{- if $.Values.ingress.alb_https_redirect }}
+          - path: /*
+            backend:
+              serviceName: ssl-redirect
+              servicePort: use-annotation
+        {{- end }}
         {{- range .paths }}
           - path: {{ . }}
             backend:
--- a/tip-wlan/resources/environments/dev-amazon-tip.yaml
+++ b/tip-wlan/resources/environments/dev-amazon-tip.yaml
@@ -64,6 +64,11 @@ opensync-mqtt-broker:
 wlan-cloud-graphql-gw:
   enabled: true
   ingress:
+    annotations:
+      kubernetes.io/ingress.class: alb
+      alb.ingress.kubernetes.io/scheme: internet-facing
+      alb.ingress.kubernetes.io/group.name: ui
+      alb.ingress.kubernetes.io/group.order: "2"
     enabled: true
     hosts:
     - host: wlan-graphql.testcluster.lab.wlan.tip.build    
@@ -76,7 +81,14 @@ wlan-cloud-static-portal:
   enabled: true
   env:
     graphql: https://wlan-graphql.testcluster.lab.wlan.tip.build
+  service:
+    type: NodePort
   ingress:
+    annotations:
+      kubernetes.io/ingress.class: alb
+      alb.ingress.kubernetes.io/scheme: internet-facing
+      alb.ingress.kubernetes.io/group.name: ui
+      alb.ingress.kubernetes.io/group.order: "1"
     hosts:
       - host: wlan-ui.testcluster.lab.wlan.tip.build
         paths: [
--- a/tip-wlan/resources/environments/dev-amazon-tip.yaml
+++ b/tip-wlan/resources/environments/dev-amazon-tip.yaml
@@ -154,6 +154,10 @@ wlan-prov-service:
     schema_repo:
       username: tip-read
       password: tip-read
+    postgres:
+      singleDataSourceUsername: tip_user
+      singleDataSourcePassword: tip_password
+      singleDataSourceSslKeyPassword: mypassword
 wlan-ssc-service:
   enabled: true
   creds:
@@ -206,10 +210,10 @@ cassandra:
   resources:
     requests:
       cpu: 500m
-      memory: 4Gi
+      memory: 3800Mi
     limits:
       cpu: 1000m
-      memory: 4Gi
+      memory: 3800Mi
   creds:
     sslKeystorePassword: mypassword
     sslTruststorePassword: mypassword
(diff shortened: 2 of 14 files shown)


tip-wlan/resources/environments/dev-amazon-tip.yaml
```diff
@@ -45,47 +40,97 @@ opensync-gw-static:
...
+    accessMode: ReadWriteOnce
     filestoreSize: 10Gi
+  tsp:
+    host: wlan-portal-service.testcluster.lab.wlan.tip.build
+  ingress:
+    enabled: true
+    alb_https_redirect: true
+    tls: []
+    annotations:
+      kubernetes.io/ingress.class: alb
```
leonidlm: We might want to use YAML Anchoring to make these annotations more DRY: https://anil.io/blog/symfony/yaml/using-variables-in-yaml-files/
With anchors, we can specify the labels once at the top of the YAML and then just assign them in the 3 or 4 places we need them defined.

tip-wlan/resources/environments/dev-amazon-tip.yaml
```diff
@@ -19,22 +19,17 @@ global:
   nsPrefix: tip
```
leonidlm, line 29: IS this `dev-amazon-tip.yaml` file is used by someone or just a template? I'm asking, because this PR adds some specific configuration values like ACM cert ids, dns names, etc... If this file is a template, we might want to create a new one that is more specific. WDYT?

tip-wlan/charts/wlan-cloud-graphql-gw/templates/ingress.yaml
```diff
@@ -32,6 +32,12 @@ spec:
     - host: {{ .host | quote }}
       http:
         paths:
+        {{- if $.Values.ingress.alb_https_redirect }}
+          - path: /*
```
leonidlm, line 36: Why do we need a feature flag "alb_https_redirect". Can we just specify these data inside the `.paths` variable? This way, we won't add AWS specific parameters to the Helm charts.

tip-wlan/resources/environments/dev-amazon-tip.yaml
eugenetaranov-opsfleet: for now it's used by other aws deployments (cicd and qa clusters)

tip-wlan/charts/wlan-cloud-graphql-gw/templates/ingress.yaml
eugenetaranov-opsfleet: that would be great option but would require to refactor all value files too (+ testing), adding just extra var helps making this change as small as possible...

tip-wlan/resources/environments/dev-amazon-tip.yaml
eugenetaranov-opsfleet: updated, please take a look: `shared.service.srv-https-annotations`

tip-wlan/resources/environments/dev-amazon-tip.yaml
eugenetaranov-opsfleet: cleaned up from deployment specific details

tip-wlan/resources/environments/dev-amazon-tip.yaml
```diff
@@ -45,47 +50,98 @@ opensync-gw-static:
...
 wlan-portal-service:
+  service:
+    type: NodePort
+    nodePort_static: false
   enabled: true
   persistence:
     enabled: true
-    storageClass: aws-efs
+    storageClass: gp2
+    accessMode: ReadWriteOnce
```
rahulsharma2020: We need ReadWriteMany accessMode to share the PV between OpensyncGW and Portal Service. As far as I remember, gp2 wont work.

tip-wlan/resources/environments/dev-amazon-tip.yaml
```diff
@@ -45,47 +50,98 @@ opensync-gw-static:
   enabled: false
 common:
   efs-provisioner:
-    enabled: true
+    enabled: false
```
rahulsharma2020: This would be needed for creating efs StorageClass, that's in turn needed for creating shared PV.

tip-wlan/charts/wlan-cloud-graphql-gw/templates/ingress.yaml
```diff
@@ -32,6 +32,12 @@ spec:
...
+            backend:
+              serviceName: ssl-redirect
+              servicePort: use-annotation
```
rahulsharma2020, line 39: In general, we define properties in values.yaml, not in the resource's yaml

tip-wlan/resources/environments/dev-amazon-tip.yaml
eugenetaranov-opsfleet: is it really needed? I was checking with @yongchen-cu and it appear to have worked okay without shared pv

tip-wlan/charts/wlan-cloud-graphql-gw/templates/ingress.yaml
eugenetaranov-opsfleet: I agree, although here if you enable alb redirection, you should not put anything different there, parametrizing here seems excessive to me.
The whole loop over Values.ingress.hosts needs to be rewritten though, paths need to be passed from values but thats not possible now due to $fullName and $svcPort being hard coded inside the loop, subject to a dedicated refactoring.
Please see also discussion here https://github.com/Telecominfraproject/wlan-cloud-helm/pull/21#discussion_r519735566

leonidlm (approved):

tip-wlan/resources/environments/dev-amazon-tip.yaml
yongchen-cu: @eugenetaranov  The PV sharing between portal and gateway is needed for things like captive service and firmware upgrade.
I only tested basic AP connectivity for your deployment, not wholesale testing with the testcluster for all SDK functionality.

eugenetaranov-opsfleet: thanks for confirmation @yongchen-cu

eugenetaranov-opsfleet: enabled

rahulsharma2020 (approved):

PR #22, opened 2020-11-16
NETEXP-485

AkshayJagadish-ne:
Added functionality for overriding external host ports (in the AWLAN table) used by the AP

### DIFF
--- a/tip-wlan/charts/opensync-gw-cloud/templates/deployment.yaml
+++ b/tip-wlan/charts/opensync-gw-cloud/templates/deployment.yaml
@@ -176,7 +176,9 @@ spec:
           env:
             {{- include "common.env" . | nindent 12 }}
             - name: OVSDB_MANAGER
-              value: {{ .Values.externalhostaddress.ovsdb }}
+              value: {{ .Values.externalhost.address.ovsdb }}
+            - name: OVSDB_EXTERNAL_PORT
+              value: "{{ .Values.externalhost.ports.ovsdb }}"
             - name: OVSDB_MANAGER_IP
               valueFrom:
                 fieldRef:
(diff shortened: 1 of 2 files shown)


tip-wlan/charts/opensync-gw-cloud/values.yaml
```diff
@@ -121,6 +121,13 @@ externalhostaddress:
   ovsdb: opensync-gw-cloud
```
rahulsharma2020: I would group the externalhost address and ports together under `externalhost`.
So something like:
```
externalhost:
   address:
      ovsdb:
      mqtt:
   ports:
      ovsdb:
      mqtt:
```

tip-wlan/charts/opensync-gw-cloud/templates/deployment.yaml
```diff
@@ -206,6 +206,10 @@ spec:
               value: {{ .Values.ethernetType.wanType }}
             - name: DEFAULT_WAN_NAME
               value: {{ .Values.ethernetType.wanName }}
+            - name: MQTT_BROKER_EXTERNAL_PORT
```
rahulsharma2020: If we can group the external variables together. Like `OVSDB_MANAGER`, `MQTT_SERVER_EXTERNAL` etc

tip-wlan/charts/opensync-gw-cloud/values.yaml
AkshayJagadish-ne: Thank you @rahulsharma2020, will make the change

rahulsharma2020 (approved):
Looks good now. Thanks @AkshayJagadish-ne

PR #25, opened 2020-11-20
resolved conflict

AkshayJagadish-ne:
Reverting commit #d113550 (WIFI-991): alb ingress, due to issues in the tip.QA deployment


tip-wlan/resources/environments/dev-amazon-tip.yaml
```diff
@@ -72,76 +60,33 @@ opensync-gw-cloud:
   persistence:
```
yongchen-cu, line 60: ovsdb and mqtt host should also be reverted back to use demo.lab.wlan.tip.build.

yongchen-cu (approved):

PR #26, opened 2020-11-23
Feature/update postgres

4c74356b41:
postgres moved to a latest bitnami chart.
namespace resource deleted for 2 reasons:
1. postgres would be deployed to default namespace with the current setup
2. this should be handled by helm using `--create-namespace` 

also: started work to unify all the certificates into common secrets to avoid certificate sprawl caused by certs folders under every chart for no good reason; added example how to run this new code under dev-local-rework values file.

a command to test this locally would be something like this:
```
helm upgrade --install tip-wlan . -f .\resources\environments\kind.yaml.local_dev -n tip --create-namespace
```


eugenetaranov-opsfleet (approved):

tip-wlan/Chart.yaml
```diff
@@ -26,54 +26,55 @@ sources:
...
+- name: kafka
+  version: 0.1.0
+  condition: kafka.enabled
+- name: postgresql
+  version: 10.1.0
+  repository: https://charts.bitnami.com/bitnami
+  condition: postgresql.enabled
+- name: cassandra
+  version: 5.5.3
+  condition: cassandra.enabled
```
rahulsharma2020, line 80: Are we missing repository here?

4c74356b41: this is a postgres PR, cassandra PR has this.

leonidlm:
@4c74356b41 for this to run, do we need to place the SSL certificates for PostgreSQL into another folder? If that's the case, we should probably update the https://github.com/Telecominfraproject/wlan-pki-cert-scripts/blob/master/copy-certs-to-helm.sh script, otherwise merging this will break the environment for everyone, no?

4c74356b41:
yeah, thats true @leonidlm, but we'd just pile all the certificates into a single place, thats the idea behind shared secrets, so you dont even need the script anymore. I mean, for now you still need the script, because many old parts still use it, but for the new code to work just copy all the certs to the new certs location

4c74356b41:
https://github.com/Telecominfraproject/wlan-cloud-helm/pull/49

PR #28, opened 2020-11-24
WIFI-752: healthchecks cloudsdk

eugenetaranov-opsfleet:
This PR improves startup speed and deduplicates cloudsdk components healthchecks.

### DIFF
--- a/tip-wlan/charts/opensync-gw-cloud/templates/deployment.yaml
+++ b/tip-wlan/charts/opensync-gw-cloud/templates/deployment.yaml
@@ -27,43 +27,13 @@ spec:
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
       initContainers:
-        - name: {{ include "common.name" . }}-mqtt-readiness
-          image: eclipse-mosquitto:latest
-          imagePullPolicy: {{ .Values.global.pullPolicy }}
-          command:
-          - sh
-          - -c
-          - |
-            mosquitto_pub -h {{ $mqtt }} -p 1883 --cafile /certs/cacert.pem --cert /certs/clientcert.pem --key /certs/clientkey.pem --insecure -t "/ap/test" -q 0 -m "CheckingMQTTAliveness"
-            status=$(echo $?)
-            echo mosquitto_pub response of the request = $status
-            counter=0
-            while [ $counter -lt 10 ] && [ $status -ne 0 ]
-            do
-              echo {{ $mqtt }} service isnt ready. Tried $counter times
-              sleep 2
-              counter=`expr $counter + 1`
-              mosquitto_pub -h {{ $mqtt }} -p 1883 --cafile /certs/cacert.pem --cert /certs/clientcert.pem --key /certs/clientkey.pem --insecure -t "/ap/test" -q 0 -m "CheckingMQTTAliveness"
-              status=$(echo $?)
-              echo mosquitto_pub response of the request = $status
-            done
-            if [ $status -eq 0 ]
-            then
-              echo {{ $mqtt }} service is ready!
-            else
-              echo {{ $mqtt }} service failed to respond after 20 secs
-              exit 1
-            fi
-          volumeMounts:
-          - mountPath: /certs/cacert.pem
-            name: certificates
-            subPath: cacert.pem
-          - mountPath: /certs/clientcert.pem
-            name: certificates
-            subPath: clientcert.pem
-          - mountPath: /certs/clientkey.pem
-            name: certificates
-            subPath: clientkey.pem
+        - name: wait-for-services
+          image: opsfleet/depends-on
+          args:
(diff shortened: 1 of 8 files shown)


leonidlm (approved):

4c74356b41 (approved):
lgtm

tip-wlan/charts/opensync-gw-cloud/templates/deployment.yaml
```diff
@@ -27,43 +27,13 @@ spec:
...
-          - mountPath: /certs/cacert.pem
-            name: certificates
-            subPath: cacert.pem
-          - mountPath: /certs/clientcert.pem
-            name: certificates
-            subPath: clientcert.pem
-          - mountPath: /certs/clientkey.pem
-            name: certificates
-            subPath: clientkey.pem
+        - name: wait-for-services
```
rahulsharma2020, line 30: If I remember correctly, the last time we tried it failed on AWS/local system since the DNS wasn't reachable. Are we using a different way now?

tip-wlan/charts/opensync-gw-cloud/templates/deployment.yaml
```diff
@@ -27,43 +27,13 @@ spec:
...
+          image: opsfleet/depends-on
```
rahulsharma2020: It's better to version the image, so we know which version is being plugged in.

tip-wlan/charts/opensync-gw-cloud/templates/deployment.yaml
```diff
@@ -27,43 +27,13 @@ spec:
...
+          args:
+          - "-service={{ .Release.Name }}-opensync-mqtt-broker"
```
rahulsharma2020, line 33: How does this list of services in arguments work?
Does the system wait for the 1st service to be ready before it goes to the next one or tries all of them simultaneously?

4c74356b41: I dont think it matters at all?

tip-wlan/charts/opensync-gw-cloud/templates/deployment.yaml
eugenetaranov-opsfleet: it failed due to k8s API being unreachable (I do not recall exactly what was the issue - wrong ip address or just something blocking access)
this PR does exactly the same and I cannot replicate it on EKS
feel free to try this on some other cluster too @rahulsharma2020

tip-wlan/charts/opensync-gw-cloud/templates/deployment.yaml
eugenetaranov-opsfleet: okay, makes sense

tip-wlan/charts/opensync-gw-cloud/templates/deployment.yaml
rahulsharma2020: Sorry @eugenetaranov-opsfleet , I don't have cycles to test it.

tip-wlan/charts/opensync-gw-cloud/templates/deployment.yaml
eugenetaranov-opsfleet: it waits till all pods in the first service come up, then proceeds to the next
https://github.com/Opsfleet/depends-on/blob/master/src/depends-on.go#L91

tip-wlan/charts/opensync-gw-cloud/templates/deployment.yaml
eugenetaranov-opsfleet: added @rahulsharma2020

tip-wlan/charts/opensync-gw-cloud/templates/deployment.yaml
```diff
@@ -27,43 +27,13 @@ spec:
...
-            name: certificates
-            subPath: cacert.pem
-          - mountPath: /certs/clientcert.pem
-            name: certificates
-            subPath: clientcert.pem
-          - mountPath: /certs/clientkey.pem
-            name: certificates
-            subPath: clientkey.pem
+        - name: wait-for-services
+          image: opsfleet/depends-on@sha256:d0eadc2b77af663d28fb24586cf8e9023ab6ce204d0d7b725299fed2e4f876e1
```
rahulsharma2020: Phew. what type of version is that :)

eugenetaranov-opsfleet: it's sha256 hash of the image, there is no up-to-date tag besides latest

rahulsharma2020: Then putting `latest` is better, until we version the image.

rahulsharma2020 (approved):

PR #29, opened 2020-11-25
WIFI-991: alb ingress

eugenetaranov-opsfleet:
Upgrade procedure:
- remove load balancers/remove route53 dns records;
- delete nginx-ingress (helm chart or deployment);
- add iam role for alb-ingress;
- install alb-ingress (switch to cloudsdk_testcluster branch in Toolsmith repo, navigate to helmfile/cloud-sdk, enable alb-ingress and run `helmfile -e amazon-testcluster -l group=system apply`);
- run helm upgrade;

PR #30, opened 2020-12-05
WIFI-990: disable hardcoded NodePort for AWS deployments

eugenetaranov-opsfleet:
Similar to https://github.com/Telecominfraproject/wlan-cloud-helm/pull/29, disabling static ports that would collide when more than one cloudsdk is deployed

### DIFF
--- a/tip-wlan/charts/common/templates/_ports.tpl
+++ b/tip-wlan/charts/common/templates/_ports.tpl
@@ -4,14 +4,17 @@
   */}}
 
 {{- define "container.dev.debugport" -}}
+  {{- if .Values.debug.enabled }}
   {{- range $index, $portid := .Values.debugPorts }}
   - name: debugport-{{ $index }}
     containerPort: {{ $portid }}
     protocol: TCP
   {{- end }}
+  {{- end }}
 {{- end -}}
 
 {{- define "service.dev.debugport" -}}
+  {{- if .Values.debug.enabled }}
   {{- range $index, $portid := .Values.debugPorts }}
   - port: {{ $portid }}
     targetPort: {{ $portid }}
@@ -20,5 +23,6 @@
     {{- if eq $.Values.service.type "NodePort" }}
     nodePort: {{ $portid }}
     {{- end }}   
+  {{- end }}
   {{- end }}
 {{- end -}}
--- a/tip-wlan/charts/wlan-cloud-graphql-gw/templates/service.yaml
+++ b/tip-wlan/charts/wlan-cloud-graphql-gw/templates/service.yaml
@@ -12,6 +12,8 @@ spec:
       targetPort: {{ .Values.service.port }}
       protocol: TCP
       name: {{ .Values.service.name }}
+      {{- if and .Values.service.nodePortStatic (eq .Values.service.type "NodePort") }}
       nodePort: {{ .Values.global.nodePortPrefix | default .Values.nodePortPrefix }}{{ .Values.service.nodePort }}
+      {{- end }}
   selector:
     {{- include "common.selectorLabels" . | nindent 4 }}
--- a/tip-wlan/charts/wlan-cloud-graphql-gw/values.yaml
+++ b/tip-wlan/charts/wlan-cloud-graphql-gw/values.yaml
@@ -51,6 +51,7 @@ service:
   port: 4000
   name: graphui
   nodePort: 23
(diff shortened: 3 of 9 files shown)


4c74356b41 (approved):
lgtm

tip-wlan/charts/common/templates/_ports.tpl
```diff
@@ -4,14 +4,17 @@
   */}}
 
 {{- define "container.dev.debugport" -}}
+  {{- if .Values.debug.enabled }}
```
rahulsharma2020, line 7: This is not correct. These are not the application debug ports. These are the debug ports that the AP developer would use to connect to the AP using the cloud, hence should not be controlled in these charts using the debug.enabled flag. See here for details: https://telecominfraproject.atlassian.net/wiki/spaces/WIFI/pages/498663439/Debug+Access+to+Customer+Equipment+using+Remote+SSH+with+PortForwardingGateway

tip-wlan/charts/common/templates/_ports.tpl
```diff
@@ -4,14 +4,17 @@
...
   {{- range $index, $portid := .Values.debugPorts }}
   - name: debugport-{{ $index }}
     containerPort: {{ $portid }}
     protocol: TCP
   {{- end }}
+  {{- end }}
 {{- end -}}
 
 {{- define "service.dev.debugport" -}}
+  {{- if .Values.debug.enabled }}
```
rahulsharma2020, line 17: Please remove this. See above comment

tip-wlan/charts/common/templates/_ports.tpl
eugenetaranov-opsfleet: Thanks for the insight @rahulsharma2020! The reason for making these ports optional is to allow multiple cloudsdk deployments. Is there a reason why making toggle for this would break the things? @rahulsharma2020 @leonidlm

rahulsharma2020: The original intent of port-forwarding gateway service is to allow remote debug of an AP. When we turn these ports off using the debug.enabled flag, the purpose of this application is jeopardized.

brennerm: Hey @rahulsharma2020, would like to find a solution for this topic together with you. Problem is that with the current configuration we won't be able to deploy multiple CloudSDK controllers onto the same Kubernetes cluster into different namespaces. Reason being the conflicting static NodePorts of the wlan-port-forwarding-gateway-service.

The above conditional expression would kind of solve this problem by only allowing one CloudSDK instance to have the debug ports enabled which isn't ideal IMO. Another quick solution I could think of is to allow defining a list of debug ports for each CloudSDK instance.

Later on I can think creating Ingresses for the debug ports which would eliminate this problem entirely.

Let me know what you think.

rahulsharma2020: Hey @brennerm, 
I think I should have renamed these ports from debugPorts to accessPointDebugPorts (or clientDebugPort) to distinguish them from the debugPort that we open to debug an application Vs these ports which would be used to debug an access point via the Cloud.

Anyway, coming back to the problem at hand. 
There are 2 problems at hand: 
1. Convert these hardcoded ports into derived ports (based on the cloud deployment in a K8s cluster)
2. Assign the values of derived ports automatically instead of statically defined by the user.
 
For '1', the one way that I can think of (which you can improvise) is to do the following:
- Instead of hardcoding these ports in values.yaml like [this](https://github.com/Telecominfraproject/wlan-cloud-helm/blob/master/tip-wlan/charts/wlan-port-forwarding-gateway-service/values.yaml#L92-L118), just define the start and end index (like startIndex = 10 and endIndex = 35) and use `extendedNodePortPrefix` (say 304) to create the NodePorts. (i.e.; change _ports.tpl -> below loops to use {{extendedNodePortPrefix}} - {{index}} to calculate the ContainerPort and NodePort respectively). 
- Then convert run.sh in PortForwardingGateway sub-chart into a configMap (as you did for others) and control the ports defined [here](https://github.com/Telecominfraproject/wlan-cloud-helm/blob/master/tip-wlan/charts/wlan-port-forwarding-gateway-service/resources/config/run.sh#L8) using override file (and using the startIndex and endIndex defined above).

For '2', perhaps we can let K8s decide on the start/endIndex and also on the extendedNodePortPrefix. However, making sure that the ports decided by K8s should lie in the list that Java (running inside the pod) can access. See [here](https://tldp.org/LDP/solrhe/Securing-Optimizing-Linux-RH-Edition-v1.3/chap6sec70.html)

brennerm:
Resolved with #34

PR #32, opened 2020-12-14
WIFI-1199 fix cassandra headless hostname

brennerm:
To be able to deploy multiple instances of the wlan-ssc-service we have to pass it the calculated value of the Cassandra headless service instead of using a hardcoded string.

### DIFF
--- a/tip-wlan/charts/wlan-ssc-service/files/cqlshrc.tip-wlan
+++ b/tip-wlan/charts/wlan-ssc-service/files/cqlshrc.tip-wlan
@@ -67,12 +67,10 @@ color = on
 ;; A version of CQL to use (this should almost never be set)
 ; version = 3.2.1
 
-
-
 [connection]
 
 ;; The host to connect to
-hostname = tip-wlan-cassandra-headless
+hostname = {{ include "cassandra.service" . }}
 
 ;; The port to connect to (9042 is the native protocol default)
 port = 9042
--- a/tip-wlan/charts/cassandra/files/cqlshrc.tip-wlan
+++ b/tip-wlan/charts/cassandra/files/cqlshrc.tip-wlan
@@ -72,7 +72,7 @@ color = on
 [connection]
 
 ;; The host to connect to
-hostname = tip-wlan-cassandra-headless
+hostname = {{ include "cassandra.service" . }}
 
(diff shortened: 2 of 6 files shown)


brennerm:
FYI @leonidlm @4c74356b41

4c74356b41:
i think there are couple of other places with that value you forgot to change?

brennerm:
Couldn't find any more. Please guide me to the correct places.

4c74356b41:
https://github.com/Telecominfraproject/wlan-cloud-helm/blob/WIFI-1199-fix-cassandra-headless-hostname/tip-wlan/charts/cassandra/resources/config/cqlshrc.default#L75

brennerm:
Seems like this file is not in use but I will adjust the `cqlshrc.tip-wlan` file next to it. Are you ok with me removing the `cqlshrc.default` file?

tip-wlan/charts/wlan-ssc-service/files/cqlshrc.tip-wlan
```diff
@@ -67,12 +67,10 @@ color = on
 ;; A version of CQL to use (this should almost never be set)
 ; version = 3.2.1
 
-
-
 [connection]
 
 ;; The host to connect to
-hostname = tip-wlan-cassandra-headless
+hostname = {{ include "cassandra.service" . }}
```
rahulsharma2020, line 73: My only doubt is if Helm will be able to over-write this variable if we are passing this file as a template to the configMap. Not tried before but looks like a good way.

rahulsharma2020 (approved):

tip-wlan/charts/wlan-ssc-service/files/cqlshrc.tip-wlan
brennerm: Not sure if I understand your comment correctly. Helm will replace this variable with the correct service name as it will copy the file into the ConfigMap. That's what we need to do to be able to achieve a multi namespace deployment.

rahulsharma2020: Yes that was my comment. Thanks, looks good.

PR #37, opened 2020-12-22
Changed mqtt version to 2.0.3 for more logs

AkshayJagadish-ne:


rahulsharma2020 (commented):
While we are it, can you also add the following lines to mosquitto.conf after `log_dest file /mosquitto/log/mosquitto.log` to increase the logging:
```
log_timestamp true
log_timestamp_format %Y-%m-%dT%H:%M:%S
log_type error
log_type warning
log_type notice
log_type information
# log_type debug
# log_type websockets
# log_type subscribe
# log_type all
connection_messages true
```

yongchen-cu (commented):
Please also make changes in the osgw-cloud deployment.yaml to synchronize to use eclipse-mosquitto:2.0.3 as well.

Please ignore my last comments, I was looking at an older verison, the dependency has been replaced with opsfleet/depends-on:latest.

rahulsharma2020 (approved):

PR #40, opened 2020-12-28
WIFI-1028 remove vendor specific default values

brennerm:
Fixes WIFI-1028
Fixes WIFI-1029
Fixes WIFI-1030


leonidlm (approved):
Looks good to me, however, before we merge this to master, please test with Roman Budilovsky that the internal values files that are used for other AWS environments won't break because of this change.

rahulsharma2020 (approved):
lgtm

PR #41, opened 2020-12-28
WIFI-1207 add example values for a multi namespace deployment

brennerm:


tip-wlan/example-values/local-multi-namespace/README.md
```diff
@@ -0,0 +1,10 @@
+# Helm values for deploying two Cloud SDK instances into separate namespaces
+
+## Usage
+
+```bash
+helm install tip-wlan-1 tip-wlan -f tip-wlan/example-values/local-multi-namespace/ns-tip-1.yaml --namespace default
+helm install tip-wlan-2 tip-wlan -f tip-wlan/example-values/local-multi-namespace/ns-tip-2.yaml --namespace default
```
leonidlm: @brennerm it's a bit confusing that we deploy the chart into the default namespace however the resources are created in another. Can we standardize on deploying the helm chart into the same namespaces as the actual resources?

brennerm: Sure let me test if that works. Usually I'd expect the resources to be created in the same namespace as the Helm chart. Is there a reason why the namespace for the resources is specified by a separate variable (global.nsPrefix) instead of just using _Release.Namspace_?

brennerm: With the current Helm chart passing a value other than `default` to the `--namespace` parameter will fail with the current error:
```
Error: create: failed to create: namespaces "tip-2" not found
```
For now I'd simply omit the `--namespace` parameter and wait for @4c74356b41's changes regarding the namespace creation.

leonidlm (approved):

PR #42, opened 2020-12-28
WIFI-1246 add default CPU/memory limits/requests

brennerm:


tip-wlan/charts/opensync-mqtt-broker/values.yaml
```diff
@@ -80,17 +80,13 @@ ingress:
...
-  # limits:
-  #   cpu: 100m
-  #   memory: 128Mi
-  # requests:
-  #   cpu: 100m
-  #   memory: 128Mi
+resources:
+  limits:
+    cpu: 100m
+    memory: 10Mi
```
leonidlm: @brennerm as discussed, in places where we need to go to such resolutions, lets be more conservative and keep to a minimum of 0.5 cpu and 128M

leonidlm (approved):

tip-wlan/charts/wlan-port-forwarding-gateway-service/values.yaml
```diff
@@ -140,17 +140,13 @@ ingress:
...
-  # limits:
-  #   cpu: 100m
-  #   memory: 128Mi
-  # requests:
-  #   cpu: 100m
-  #   memory: 128Mi
+resources:
+  limits:
+    cpu: 500m
+    memory: 350Mi
```
rahulsharma2020, line 146: What's the basis of cpu and memory numbers. Experimentally derived?

tip-wlan/charts/opensync-gw-cloud/values.yaml
```diff
@@ -166,17 +166,13 @@ ingress:
...
-  # limits:
-  #   cpu: 100m
-  #   memory: 128Mi
-  # requests:
-  #   cpu: 100m
-  #   memory: 128Mi
+resources:
+  limits:
+    cpu: 500m
+    memory: 750Mi
```
rahulsharma2020, line 172: I am sure that opensync-gw will hit this limit soon'ish. I remember @cbuschfb sent a message in the wlan-cloud-platform regarding running with many Simulated APs (see here: https://telecominfraproject.slack.com/archives/G019WQETNG3/p1607697089088200). Perhaps would be good to know the limits Opensync-GW pod reached during that exercise and then using it to benchmark OpensyncGW pod limits (based on the number of APs one of the instances can manage).

tip-wlan/charts/wlan-port-forwarding-gateway-service/values.yaml
brennerm: All values are based off of the metrics collected and displayed at https://grafana.lab.wlan.tip.build

tip-wlan/charts/opensync-gw-cloud/values.yaml
brennerm: My goal with these default values is to allow Kubernetes to properly schedule a Cloud SDK deployment + a certain (we have to find out what certain means in our case) amount of additional capacity for some load. With enough APs generating load you'll always be able to exceed these limits. That's why they should be overwritten when it's necessary. Hope that makes sense.

tip-wlan/charts/wlan-port-forwarding-gateway-service/values.yaml
rahulsharma2020: Thanks. I only see data from Prometheus in this Grafana cluster and only for a few pods (Cassandra, Postgres, kafka, nginx and mosquitto), hence the question for other pods.
Were there APs managed by this deployment when the limits were observed?

tip-wlan/charts/opensync-gw-cloud/values.yaml
rahulsharma2020: Yes sure. I was thinking a bit in the future I suppose, on figuring out the # of supported APs/Opensync-gw-pod before the limits set here are reached.

tip-wlan/charts/wlan-port-forwarding-gateway-service/values.yaml
brennerm: Mmh, that's weird. I can see all these pods.
image

Not sure, but definitely there weren't too many APs connected.

tip-wlan/charts/opensync-gw-cloud/values.yaml
brennerm: Feel free to make other suggestions. I'm just missing the data to make better predictions for these values.

tip-wlan/charts/wlan-port-forwarding-gateway-service/values.yaml
rahulsharma2020: Weird, I still don't see it. Can you send me the link where you see the pods.
Also, most of _our_ (portal, prov, ssc, spc, opensync-gw etc) pods don't have prometheus agent running, hence you won't get the CPU/memory data from them.

tip-wlan/charts/opensync-gw-cloud/values.yaml
rahulsharma2020: Sure. I think we can continue with these for now and then update them to determine the right #s based on # of APs after experimental tests.

tip-wlan/charts/wlan-port-forwarding-gateway-service/values.yaml
brennerm: Try if that works for you: https://grafana.lab.wlan.tip.build/d/85a562078cdf77779eaa1add43ccec1e/kubernetes-compute-resources-namespace-pods?orgId=1&refresh=10s&var-datasource=Prometheus&var-cluster=&var-namespace=tip

4c74356b41: works for me

rahulsharma2020: Thanks, I see it now. My filters were not set.

rahulsharma2020 (approved):

PR #45, opened 2021-01-14
decrease cpu requests + Cassandra startup fix

brennerm:
The CPU requests I set are way to high and prevent a simple single node deployment.

Additionally Cassandra seems to require a lot memory on startup which is also fixed in this PR.


4c74356b41 (approved):
lgtm

PR #49, opened 2021-01-19
Feature/thirdparties

4c74356b41:
postgres, kafka, cassandra moved to a latest bitnami charts and are being pulled during the install\upgrade.
namespace resource deleted for 2 reasons:
1. thirdparties would be deployed to default namespace with the current setup instead of the tip-wlan
2. this should be handled by helm using `--create-namespace` 

added example how to run this new code under dev-local-thirdparties values file, a command to test this locally would be something like this:
```
helm upgrade --install tip-wlan . -f .\resources\environments\dev-local.yaml -n tip --create-namespace
```

added an example of how to migrate data with existing cloud sdk deployments under migration.yaml. essentially cassandra and postgres will just reuse existing pvc and that appears to be working. I believe kafka and mqtt data doesn't need to be migrated - hence I didnt test that.

also: started work to unify all the certificates into common secrets to avoid certificate sprawl caused by certs folders under every chart for no good reason, an example of how to copy the certs to the new location
https://github.com/Telecominfraproject/wlan-pki-cert-scripts/blob/local/dev/copy-certs-to-helm.sh#L42

short summary of the migration procedure:

1. `helm uninstall -n default tip-wlan` (this doesn't delete PVC)
2. checkout latest version of this branch (or master if its merged already)
3. Update your values file (used to deploy this chart) with the values from migration.yaml to support thirdparties and preserve existing cassandra\postgres data. You need to override existing values for postgres\cassandra\kafka and add the `creds` property and sub properties to the `global` values (it used to be under cassandra).
4. `rm -rf charts/cassandra charts/kafka charts/postgresql`
5. `find . -regextype posix-extended -regex '.+(jks|pem|key|pkcs12|p12)$' -exec cp "{}" resources/certs/ \;`
6. `helm dependency update`
7. run the updated cert copy script
8. `helm upgrade --install -n tip --create-namespace -f .\resources\environments\your_values_with_fixes.yaml`


brennerm:
@4c74356b41 The deployment seems to fail. Can you check that?

4c74356b41:
yeah, deployment will fail, since its not able to handle new way of deploying the chart (needs `helm dependency update`, new values file as well).

brennerm:
Ahh ok, so feel free to add it to the workflow file.

brennerm:
Think we should also add a hint to the README that instructs the user to execute `helm dependency update` before `helm install`. Don't think we should expect everyone to know that.

brennerm:
@4c74356b41 When deploying your branch from scratch (=no previous CloudSDK installation) the Cassandra and PostgreSQL pod are stuck in the pending state cause they are missing their corresponding PV:
`0/1 nodes are available: 1 pod has unbound immediate PersistentVolumeClaims`

4c74356b41:
cant debug that, unless you show how you install that, but I assume your storage classes are wrong

brennerm:
> cant debug that, unless you show how you install that, but I assume your storage classes are wrong
Here's what I did:
1. start minikube with 4 CPUs and 16 GB of RAM
2. check out wlan-cloud-helm repo, feature/thirdparties branch
3. generate certs in wlan-pki-cert-scripts using local/dev branch
4. copy over certs to wlan-cloud-helm folder
5. execute `helm upgrade --install tip-wlan . -f resources/environments/dev-local-thirdparties.yaml -n tip --create-namespace`

4c74356b41:
yeah, that's pretty obvious, like I said storage classes, reverted them to match what minikube has. sorry, my bad.

leonidlm:
@4c74356b41 are you planning to have a step-by-step documentation for how to perform the migration from the old charts version to your setup? I saw the migration.yaml in the PR but I'm sure people (me included) will benefit for a details step-by-step guide for how to move to the newer version.

4c74356b41:
yes, I'll update the PR with this information tomorrow.

leonidlm:
@4c74356b41 notes from our meeting today:

- We need to update the values in the `environments` folder to support the breaking changes & test with relevant people. Maybe an opportunity for a cleanup or unused env files.

brennerm:
@4c74356b41 Was able to successfully deploy a Cloud SDK instance, access the UI and connect my AP with your branches. :+1: 

Now I'd like to test the migration process. What's the best way to check if the migration was successful

brennerm:
FYI I see a lot of these errors in Kafka which seems unusual:
```
tip-wlan-kafka-0 kafka [2021-01-21 11:11:31,407] WARN [SocketServer brokerId=0] Unexpected error from /172.17.0.6; closing connection (org.apache.kafka.common.network.Selector)
tip-wlan-kafka-0 kafka org.apache.kafka.common.network.InvalidReceiveException: Invalid receive (size = 369296129 larger than 104857600)
tip-wlan-kafka-0 kafka  at org.apache.kafka.common.network.NetworkReceive.readFrom(NetworkReceive.java:105)
tip-wlan-kafka-0 kafka  at org.apache.kafka.common.network.KafkaChannel.receive(KafkaChannel.java:447)
tip-wlan-kafka-0 kafka  at org.apache.kafka.common.network.KafkaChannel.read(KafkaChannel.java:397)
tip-wlan-kafka-0 kafka  at org.apache.kafka.common.network.Selector.attemptRead(Selector.java:678)
tip-wlan-kafka-0 kafka  at org.apache.kafka.common.network.Selector.pollSelectionKeys(Selector.java:580)
tip-wlan-kafka-0 kafka  at org.apache.kafka.common.network.Selector.poll(Selector.java:485)
tip-wlan-kafka-0 kafka  at kafka.network.Processor.poll(SocketServer.scala:913)
tip-wlan-kafka-0 kafka  at kafka.network.Processor.run(SocketServer.scala:816)
tip-wlan-kafka-0 kafka  at java.base/java.lang.Thread.run(Thread.java:834)
```

4c74356b41:
@brennerm create a new AP profile in cloud sdk and perform the migration. if it is successful you should see it in the new cloud sdk

tip-wlan/templates/kafka-dependencies.yaml
```diff
@@ -0,0 +1,97 @@
...
+  name: {{ .Release.Namespace }}-wlan-kafka-config-{{ .Release.Revision }}
+  namespace: {{ .Release.Namespace }}
+spec:
+  backoffLimit: 10
+  template:
+    spec:
+      restartPolicy: OnFailure
+      initContainers:
+      - name: depends-on
+        image: confluentinc/cp-kafka:5.0.1
```
leonidlm: @4c74356b41 should we use Bitnami's image here instead because we are deploying Bitnami Kafka

leonidlm: @4c74356b41 do we want to parametrize the depends-on Docker image version and also the port number below (9092)

rahulsharma2020 (approved):

.github/workflows/testing.yml
```diff
@@ -26,6 +27,7 @@ jobs:
       with:
         path: wlan-pki-cert-scripts
         repository: Telecominfraproject/wlan-pki-cert-scripts
+        ref: local/dev
```
leonidlm: @4c74356b41 I think we can merge this to master as part of this merge?

.github/workflows/testing.yml
```diff
@@ -40,7 +42,7 @@ jobs:
 
     - name: Generate Helm values file
       run: |
-        ./Toolsmith/helm-values/aws-cicd-pr-deployment.yaml.sh ${{ env.PR_NUMBER }} > pr-deployment.yaml
+        ./Toolsmith/helm-values/aws-cicd-testing-pr-deployment.yaml.sh ${{ env.PR_NUMBER }} > pr-deployment.yaml
```
leonidlm, line 44: Why do you think we need a separate YAML file? I guess you didn't touch the original because you didn't want to break the build... we just need to remember to remove the original then once this PR is merged.

leonidlm (approved):
Looks good to me! @4c74356b41

.github/workflows/testing.yml
4c74356b41: i'd then need to address the comments on that one, maybe its better to do it directly after?

.github/workflows/testing.yml
4c74356b41: yes, this would break the build for other branches, unfortunately. yes, we need to fix this post merge.

.github/workflows/testing.yml
leonidlm: @4c74356b41 ok, lets do it after we merge this PR and then fix back.

tip-wlan/resources/environments/dev-local.yaml
```diff
@@ -52,22 +63,6 @@ opensync-gw-cloud:
     enabled: true
   filestore:
     url: "https://wlan-filestore.wlan.local"
-  scalability:
```
leonidlm, line 55: @4c74356b41 wondering why this section was deleted?

.github/workflows/testing.yml
leonidlm: @4c74356b41 now that I think about this... You won't be able to merge this PR without merging the local/dev branch in the PKI script, because otherwise, you will break the cloudsdk deployment for everyone who is using the PKI script with the master.

brennerm:
@4c74356b41 Had to delete the obsolete chart directories (`rm -rf charts/cassandra charts/kafka charts/postgresql`) after switching to the new branch before I was able to successfully execute `helm dependency update`. Please add that step to the migration guide.

README.md
```diff
@@ -1,16 +1,66 @@
...
+```
+rm -rf tip-wlan/charts/cassandra tip-wlan/charts/kafka tip-wlan/charts/postgresql
+```
+6. Pull 3rd party subcharts:
+```
+helm dependency update tip-wlan
+```
+7. Perform Helm upgrade:
+```
+helm upgrade --install tip-wlan tip-wlan/ --namespace tip --create-namespace -f .\resources\environments\your_values_with_fixes.yaml
```
brennerm: Think we should default to forward slashes here.

README.md
```diff
@@ -1,16 +1,66 @@
...
+3. Update your values file that you used for deploying the original release with the values from `migration.yaml` to preserve existing cassandra\postgres data (or skip that step and use the second upgrade command mentioned in #7)
+4. If you want to preserve the PKI certificates from the original Helm installation, copy them to a new location using the command below (or checkout the latest wlan-pki-cert-script repo and use `copy-certs-to-helm.sh %path_to_new_helm_code%` to generate new self-signed keys):
+```
+find . -regextype posix-extended -regex '.+(jks|pem|key|pkcs12|p12)$' -exec cp "{}" tip-wlan/resources/certs/ \;
+```
+5. Remove the old charts from the helm directory, so that the upgrade command can successfully pull new chart depedencies:
+```
+rm -rf tip-wlan/charts/cassandra tip-wlan/charts/kafka tip-wlan/charts/postgresql
+```
+6. Pull 3rd party subcharts:
```
brennerm, line 35: We should mention that this command should be executed in the tip-wlan folder.

tip-wlan/values.yaml
```diff
@@ -1,31 +1,46 @@
-# Default values for Tip-Wlan
-# This is a YAML-formatted file.
-# Declare variables to be passed into your templates.
+shared:
+  scalability: &jvm-options
```
brennerm, line 2: This block isn't used anywhere or do I miss something here?

4c74356b41: its not for now, but the idea is to have everything in values and override from resources\environments if\when necessary.

PR #50, opened 2021-01-19
WIFI-1287

AkshayJagadish-ne:
*Removed run.sh as it overrides the run.sh from the back-end
https://github.com/Telecominfraproject/wlan-cloud-services/blob/master/port-forwarding-gateway-docker/src/main/docker/app/run.sh

*created two env variables EXT_PORT_RANGE_START and
EXT_PORT_RANGE_END with values {{ include "apDebugPortsStart" . }} and
{{ include "apDebugPortsStart" . }} respectively


rahulsharma2020 (commented):
@AkshayJagadish-ne: run.sh is maintained here because of the following line: https://github.com/Telecominfraproject/wlan-cloud-helm/blob/master/tip-wlan/charts/wlan-port-forwarding-gateway-service/files/run.sh#L8, which refers to the values coming from Helm values.yaml. 
If you want to use the run.sh from backend, we need to find a way to pass in sysctl parameter to it with these port-ranges

AkshayJagadish-ne:
> @AkshayJagadish-ne: run.sh is maintained here because of the following line: https://github.com/Telecominfraproject/wlan-cloud-helm/blob/master/tip-wlan/charts/wlan-port-forwarding-gateway-service/files/run.sh#L8, which refers to the values coming from Helm values.yaml.
> If you want to use the run.sh from backend, we need to find a way to pass in sysctl parameter to it with these port-ranges

Thank you @rahulsharma2020, the back-end uses LOCAL_PORT_RANGE_START/END https://github.com/Telecominfraproject/wlan-cloud-services/blob/bde975131139fa0cc03a23bd4154b3c92fa745ab/port-forwarding-gateway-docker/src/main/docker/app/run.sh#L10, 

so I have added these two env variables and pointed them to {{ include "apDebugPortsStart" . }} and
{{ include "apDebugPortsStart" . }} respectively

tip-wlan/charts/wlan-port-forwarding-gateway-service/templates/deployment.yaml
```diff
@@ -71,6 +71,10 @@ spec:
               value: {{ .Values.externallyVisible.host }}
             - name: PF_GATEWAY_RELEASE_EXT_PORT
               value: "{{ .Values.externallyVisible.port }}"
+            - name: EXT_PORT_RANGE_START
+              value: "{{ include "apDebugPortsStart" . }}"
+            - name: EXT_PORT_RANGE_END
+              value: "{{ include "apDebugPortsEnd" . }}"
```
rahulsharma2020: I think this should be `{{ sub (include "apDebugPortsEnd" . | atoi) 1 }}` to subtract 1 from the last value. See the value used in the now removed run.sh

tip-wlan/charts/wlan-port-forwarding-gateway-service/files/README.txt
```diff
@@ -0,0 +1,3 @@
+#This README prevents
```
rahulsharma2020: I don't think we need this file or the folder, if it's not being used in the charts. We should update the configMap to remove reference to the `files` folder.

tip-wlan/charts/wlan-port-forwarding-gateway-service/templates/deployment.yaml
```diff
@@ -71,6 +71,10 @@ spec:
               value: {{ .Values.externallyVisible.host }}
             - name: PF_GATEWAY_RELEASE_EXT_PORT
               value: "{{ .Values.externallyVisible.port }}"
+            - name: EXT_PORT_RANGE_START
```
brennerm, line 74: In your comment you mention that the env variable is called LOCAL_PORT_RANGE_START. Here you are referring to EXT_PORT_RANGE_START. Which one is correct?

tip-wlan/charts/wlan-port-forwarding-gateway-service/templates/deployment.yaml
```diff
@@ -71,6 +71,10 @@ spec:
...
+              value: "{{ include "apDebugPortsStart" . }}"
+            - name: EXT_PORT_RANGE_END
```
brennerm, line 76: In your comment you mention that the env variable is called LOCAL_PORT_RANGE_END. Here you are referring to EXT_PORT_RANGE_END. Which one is correct?

AkshayJagadish-ne: @brennerm , I was referring to EXT_PORT_RANGE_END, not LOCAL_PORT_RANGE_END. I will make the correction in the description. Thank you!

brennerm (approved):

rahulsharma2020 (approved):

dmitry-toptygin-connectus-ai (approved):

PR #51, opened 2021-01-25
WIFI-1172 add JMX to Prometheus PoC

brennerm:
This is a PoC to show how to expose JMX metrics via Prometheus.

In the "real" implementation I would additionally:
- make the JMX exporter agent version configurable
- introduce a flag to enable/disable integrating the JMX exporter agent
- introduce a flag to enable/disable creation of the PodMonitor resource 
- introduce templates for everything to make it reusable for all services


brennerm:
A resulting example metric can be viewed [here](https://grafana.lab.wlan.tip.build/explore?orgId=1&left=%5B%221611584268216%22,%221611587085427%22,%22Prometheus%22,%7B%22expr%22:%22jvm_memory_bytes_used%22%7D,%7B%22mode%22:%22Metrics%22%7D,%7B%22ui%22:%5Btrue,true,true,%22none%22%5D%7D%5D).

tip-wlan/charts/opensync-gw-cloud/templates/deployment.yaml
```diff
@@ -34,6 +34,17 @@ spec:
...
           - "-service={{ .Release.Name }}-wlan-ssc-service"
           - -check_interval=5
+        - name: download-jmx-prometheus-exporter
+          image: alpine:latest
+          command:
+          - wget
+          args:
+          - -P
+          - /jmx-prometheus-exporter-dir
+          - https://repo1.maven.org/maven2/io/prometheus/jmx/jmx_prometheus_javaagent/0.14.0/jmx_prometheus_javaagent-0.14.0.jar
```
dmitry-toptygin-connectus-ai: Better to point this to a mirror in the JFrog, instead of the maven central. I know, it's a prototype, but there's nothing more permanent than a temporary thing :)

tip-wlan/charts/opensync-gw-cloud/templates/deployment.yaml
```diff
@@ -142,7 +153,7 @@ spec:
             - name: tip_wlan_maxHttpThreads
               value: "{{ .Values.scalability.tip_wlan_maxHttpThreads }}"
             - name: JVM_MEM_OPTIONS
-              value: "{{ .Values.scalability.JVM_MEM_OPTIONS }}"
+              value: "{{ .Values.scalability.JVM_MEM_OPTIONS }} -javaagent:/jmx-prometheus-exporter-dir/jmx_prometheus_javaagent-0.14.0.jar=9404:/app/jmx-prometheus-config.yml"
```
dmitry-toptygin-connectus-ai: what about security for the jmx_prometheus_javaagent agent? for both access control and transport (to protect against traffic sniffing/injecting). can we configure TLS with certificate-based auth?

brennerm: Not sure if I understand your issue here. When running the exporter as a Java agent there is no network communication at all. Everything goes through altering/reading bytecode within the JVM. Let me know if I miss something here.

dmitry-toptygin-connectus-ai: So, the agent will not be accepting network connections on port 9404?
I mean connections from things like JMXConsole, which can access any MBean and call methods on them that not only can read things, but to also change the limits on pool sizes, trigger garbage collection, of invoke arbitrary methods on arbitrary java classes.
Maybe I'm missing something and the jmx_prometheus_javaagent does not accept network connections from the outside world? Does it push the data to the prometheus instead?

dmitry-toptygin-connectus-ai: Try connecting jconsole to port 9404. I'm curious what happens.
https://docs.oracle.com/javase/8/docs/technotes/guides/management/jconsole.html
...
I see now, the agent exposes port that provides http access to it, hopefully not allowing full JMX control.

dmitry-toptygin-connectus-ai: https://github.com/prometheus/jmx_exporter 
"This exporter is intended to be run as a Java Agent, exposing a HTTP server and serving metrics of the local JVM. "
To me this feels weird - to run an unprotected HTTP server that is directly embedded as an agent into the bowels of the server process. Even if it does not expose full capabilities of JMX, it's just asking for trouble.

brennerm: The port 9404 is for the HTTP server so a JMX console will not be able to connect to it.

IMO you are a little overcautious here. We are talking about a read only HTTP interface with non confidential (at least for now) information that is never being exposed outside of the cluster.

leonidlm: Just to add to the discussion here - read-only HTTP endpoints are pretty common in the Kubernetes-monitoring space. Prometheus scraping works by periodically pulling HTTP endpoints exposed by the application for metrics reporting purposes. As Prometheus' security model suggests (https://prometheus.io/docs/operating/security/) - metrics aren't considered secrets, and despite the fact that Prometheus supports TLS for scraping, many available community exporters rarely implement HTTPs support.

I suggest we go forward with the implementation, and if needed, add the TLS support in the future since the `jmx_exporter` actually supports that.

dmitry-toptygin-connectus-ai: My personal opinion after reading the page quoted (https://prometheus.io/docs/operating/security/) is that prometheus security model is a joke. I would strongly suggest to use TLS wherever it is available. Nobody knows what turns the project takes in the future, nobody knows if at some point the deployment takes a path of having several clusters on public untrusted cloud - for variour reasions - money, politics, scalability requirements, etc. If you select a path of least resistance now (unencrypted internal traffic) - you may put the project into a corner in the future.

brennerm: I don't say that mTLS is a no-go. But it shouldn't be the default and available as an addon for setups that require increased safety measures. It adds complexity that isn't necessary for dev envs and keeps users from getting started with CloudSDK.

dmitry-toptygin-connectus-ai: That decision should be made by @cbuschfb and @Dorongiv.

dmitry-toptygin-connectus-ai: If we put on hold my comments about mTLS/TLS, how far away are we from making this prototype a reality?
This feature can always be configured on/off, and it provides a lot of value for the scalability testing.

brennerm: It's not too big of a deal. The configuration will be the same for all services + some templating to reduce duplications.

brennerm:
Here's where I'm at right now. Added the JMX Prometheus exporter to all JVM services and introduced reusable templates.
Result:
image

The whole functionality is disabled by default right now and can be enabled with the flags `global.monitoring.enableJmxPrometheusMetrics` and `global.monitoring.enablePrometheusPodMonitors`

dmitry-toptygin-connectus-ai:
I propose we go with this - merge it into master, if nobody objects. 
It is marked as a "Draft PR" right now, so it cannot be merged into master yet - needs to be converted to a regular one.

leonidlm (approved):
Looks good!

dmitry-toptygin-connectus-ai (approved):

tip-wlan/charts/opensync-gw-cloud/templates/deployment.yaml
```diff
@@ -187,6 +190,7 @@ spec:
               containerPort: {{ .Values.service.port5 }}
               protocol: TCP
             {{- end }}
+            {{- include "jmxPrometheus.port" . | nindent 12 }}
```
rahulsharma2020, line 193: Thanks @brennerm for the PR. Looks good.
Small concern is that we cannot use it in MicroK8s env (since we are using same port number for all services and metalLB would complain in it's current form (I can see possible clash for OpensyncGw and PortalService LoadBalancer services)).
But it's good for simplicity.

tip-wlan/templates/_prometheus-jmx.tpl
```diff
@@ -0,0 +1,93 @@
...
+{{- end -}}
+
+{{- define "jmxPrometheus.portNumber" -}}
+9404
+{{- end -}}
+
+{{- define "jmxPrometheus.initContainer" -}}
+{{- if .Values.global.monitoring.enableJmxPrometheusMetrics -}}
+- name: download-jmx-prometheus-exporter
+  image: alpine:latest
```
rahulsharma2020, line 20: Would be good to use a specific version as image tag

tip-wlan/charts/opensync-gw-cloud/templates/deployment.yaml
brennerm: Not sure if I understand your concern correctly. I haven't introduced any new service ports. The only thing I added are Pod ports that are exposed at the Pod internal cluster IP and thus do not collide.

rahulsharma2020: Oh yeah, my bad :)

rahulsharma2020 (approved):

brennerm:
@dmitry-toptygin-connectus-ai I'd be glad if you could let the PR creator decide when to merge. I still had local changes that I'd like to have included in this PR.

dmitry-toptygin-connectus-ai:
@brennerm Sorry, if I jumped the gun on this one. "in my defence the moon was full and I was left unsupervised" :) 

My reasoning for the merge was: it has been in review for a while, you said it is working and provided a screenshot proving so, Leo approved it, Rahul approved it, I did not see anything obviously wrong with it, and internally we are going through the scalability testing where we would be using this functionality, so I did not want to re-invent the wheel. The branch can be restored and you can continue submitting your changes to this PR.

Process-wise, I was operating under the assumption that as soon as critical mass of approvers is reached, the PR can be merged. I can stay away from merging this repo altogether to not get into this kind of situation in the future.

PR #53, opened 2021-01-29
fixing docker secret and fixing kafka topics

4c74356b41:
docker secret generation fixes
kafka topics fixes

### DIFF
--- a/tip-wlan/values.yaml
+++ b/tip-wlan/values.yaml
@@ -37,10 +37,12 @@ global:
   nodePortPrefix: 302
   nodePortPrefixExt: 304
 
-  repository: tip-tip-wlan-cloud-docker-repo.jfrog.io
+  repository:
+    registry: tip-tip-wlan-cloud-docker-repo.jfrog.io
+    username: tip-read
+    password: tip-read
   # image pull policy
   pullPolicy: Always
-
 
   # flag to enable debugging - application support required
   debugEnabled: false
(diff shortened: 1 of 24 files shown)


tip-wlan/values.yaml
```diff
@@ -37,11 +37,13 @@ global:
   nodePortPrefix: 302
   nodePortPrefixExt: 304
 
-  repository: tip-tip-wlan-cloud-docker-repo.jfrog.io
+  repository:
+    registry: tip-tip-wlan-cloud-docker-repo.jfrog.io
+    username: tip-read
+    password: tip-read
```
brennerm, line 43: Please don't tell me that this is the actual password. :sweat_smile:

brennerm:
These are some very nice improvements. :+1: 

Do you also want to fix the example values in tip-wlan/example-values/local-multi-namespace? Guess the README also needs adjustment with the changes to the namespace handling.

tip-wlan/values.yaml
4c74356b41: it is for the read-only account (as far as I'm aware). Its available in plain text in many other places, so I dont think its an issue. I've verified in jFrog that it can only read stuff.

brennerm: Don't get me wrong, I'm not concerning about you adding it here. It's just the password itself. :smile:

brennerm (approved):

brennerm (changes requested):
~~The Cassandra dependencies need fixing as well, e.g. https://github.com/Telecominfraproject/wlan-cloud-helm/blob/master/tip-wlan/templates/cassandra-dependencies.yaml#L10~~

Mistake on my side.

brennerm (approved):

PR #54, opened 2021-01-30
WIFI-1478 - all credentials moved to globals

4c74356b41:
values files changes:
removed `dockerRegistrySecret` as obsolete
removed `annotations` as obsolete
moved scalabilty completely to base values.yaml to not repeat this in values files
updated multi-namespace examples to support latest changes

other changes:
centralized secrets in helm values and moved certificates secrets to parent chart
added startup probes to prevent container restarts during startup process
removed `.Values.creds.enabled` conditions, because they do not make sense, none of the services using those conditions would properly work without those


CHANGELOG.md
```diff
@@ -8,7 +8,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
```
brennerm: Think we should add a Changelog entry for these changes. While you are at it you also add one for the integration of the third party Helm charts.

4c74356b41: i never used integration so i'd prefer to keep it out (as I have no idea how to test it). we can raise another PR for integration thingie. as for the changelog - sure, do you have something in mind?

brennerm: Not sure if you understood me correct here. I just think we are missing a changelog entry for using the third party Helm charts instead of the custom ones we had before.

My proposal:
```
### Added
- Bitnami Helm charts for Kafka, Cassandra and PostgreSQL

### Changed
- extract common credentials into a unified place

### Removed
- custom Helm charts for Kafka, Cassandra and PostgreSQL
```

brennerm (approved):

CHANGELOG.md
4c74356b41: the pr kinda describes all the changes? do you want to put that into the readme or something?

brennerm: No I want to put it into the CHANGELOG.md file so that everyone can see what has changed since the 0.4 release.

brennerm: Think @leonidlm wanted to use the [keepachangelog format](https://keepachangelog.com/en/1.0.0/). Additionally I think all of these changes need to be added to the Unreleased section as 1.0.0 has not yet been released.

tip-wlan/charts/opensync-gw-static/templates/deployment.yaml
```diff
@@ -47,7 +47,12 @@ spec:
             failureThreshold: {{ .Values.probes.readinessProbe.failureThreshold }}
             periodSeconds: {{ .Values.probes.readinessProbe.periodSeconds }}
             successThreshold: {{ .Values.probes.readinessProbe.successThreshold }}
-          {{- end }}       
+          startupProbe:
+            tcpSocket:
+              port: {{ .Values.service.port2 }}
```
leonidlm, line 52: Can we have a more descriptive names than `port1` or `port2` ?

tip-wlan/charts/wlan-integrated-cloud-component-service/templates/secret.yaml
```diff
@@ -22,9 +22,7 @@ metadata:
     heritage: {{ .Release.Service | quote }}
 type: Opaque
 data:
-  {{- if .Values.creds.enabled }}
-  postgresql-password: {{ .Values.creds.postgres.password | b64enc | quote }}
-  schema-repo-user: {{ .Values.creds.schema_repo.username | b64enc | quote }}
-  schema-repo-password: {{ .Values.creds.schema_repo.password | b64enc | quote }}
-  {{- end }}
+  postgresql-password: {{ .Values.global.postgres.password | b64enc | quote }}
```
leonidlm, line 25: Why have we removed the `if ... creds.enabled` clause?

tip-wlan/charts/wlan-port-forwarding-gateway-service/templates/secret.yaml
```diff
@@ -19,6 +10,4 @@ metadata:
     heritage: {{ .Release.Service | quote }}
 type: Opaque
 data:
-  {{- if .Values.creds.enabled }}
-  websocketSessionTokenEncKey: {{ .Values.creds.websocketSessionTokenEncKey | b64enc | quote }}
-  {{- end }}
\ No newline at end of file
+  websocketSessionTokenEncKey: {{ .Values.creds.websocketSessionTokenEncKey | b64enc | quote }}
```
leonidlm, line 13: Same comment here as before - why we remove the if clause?

tip-wlan/charts/wlan-prov-service/templates/configmap.yaml
```diff
@@ -4,16 +4,16 @@ metadata:
   name: {{ include "common.fullname" . }}-log-config
   namespace: {{ include "common.namespace" . }}
 data:
-{{ tpl (.Files.Glob "resources/config/*").AsConfig . | indent 2 }}
+{{ (.Files.Glob "resources/config/logback.xml").AsConfig | indent 2 }}
   datasource.properties: |-
     singleDataSource.url=jdbc:postgresql://{{- include "postgresql.service" . -}}:5432/prov_db
-    singleDataSource.username={{ .Values.creds.postgres.singleDataSourceUsername }}
-    singleDataSource.password={{ .Values.creds.postgres.singleDataSourcePassword }}
+    singleDataSource.username={{ .Values.global.postgres.singleDataSource.username }}
```
leonidlm, line 10: When we use global values, we should probably provide an option to override them locally. A pattern used across many Helm charts goes something like: `{{ .Values.global.postgres.singleDataSource.username | default .Values.postgres.singleDataSource.username }}`

tip-wlan/charts/wlan-prov-service/templates/deployment.yaml
```diff
@@ -169,6 +169,13 @@ spec:
             failureThreshold: {{ .Values.probes.readinessProbe.failureThreshold }}
             periodSeconds: {{ .Values.probes.readinessProbe.periodSeconds }}
             successThreshold: {{ .Values.probes.readinessProbe.successThreshold }}
+          startupProbe:
+            httpGet:
+              path: /ping
+              port: {{ .Values.service.port2 }}
+              scheme: {{ .Values.probes.readinessProbe.scheme }}
+            failureThreshold: 500
```
leonidlm: The probe parameters shouldn't be hard coded into the template. We can move them to the values of that chart with reasonable defaults.

tip-wlan/charts/wlan-ssc-service/files/cqlshrc.tip-wlan
```diff
@@ -19,12 +19,11 @@
 
 [authentication]
 ;; If Cassandra has auth enabled, fill out these options
-username = cassandra 
+username = cassandra
```
leonidlm, line 22: Shouldn't this be a template instead of hardcoded user/pass?

tip-wlan/example-values/local-multi-namespace/README.md
```diff
@@ -3,8 +3,8 @@
 ## Usage
 
 ```bash
-helm install tip-wlan-1 tip-wlan -f tip-wlan/example-values/local-multi-namespace/ns-tip-1.yaml
-helm install tip-wlan-2 tip-wlan -f tip-wlan/example-values/local-multi-namespace/ns-tip-2.yaml
+helm install tip-wlan-1 tip-wlan -f tip-wlan/example-values/local-multi-namespace/ns-tip-1.yaml --namespace tip-wlan-1 --create-namespace
```
leonidlm: Maybe we should update the docs to use `helm upgrade --install` instead of the `helm install` command? The upgrade command is idempotent which is better.

tip-wlan/resources/certs/README.md
```diff
@@ -0,0 +1,2 @@
+Contains certs needed for this service to start.
```
leonidlm: NICE!!!

CHANGELOG.md
```diff
@@ -7,7 +7,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-
+## [1.0.1]()
+
+values files changes:
```
leonidlm: Changelog shouldn't include the complete description of each PR - we have the PR description and JIRA (assuming it will be open at some point) for people who would like to dive deeper into the change details.

Instead, Changelog entries should have one or two sentences max per each PR with a descriptive title that gives a quick hint of the change introduced in each PR.

Although we aren't following the Kubernetes' changelog format, but here's an example for the level of detail we want to keep in the Changelog: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.18.md#feature-1

Please also include a link to the PR at the end of each entry. For example:

```
### Added
- Bitnami Helm charts for Kafka, Cassandra and PostgreSQL ([#49](https://github.com/Telecominfraproject/wlan-cloud-helm/pull/49))
```

leonidlm: While I'm at it, we also need to add a changelog entry for: https://github.com/Telecominfraproject/wlan-cloud-helm/pull/53

leonidlm (changes requested):

tip-wlan/charts/wlan-integrated-cloud-component-service/templates/secret.yaml
4c74356b41: yeah, mentioned this in the pr notes, they don't make any sense as the service won't work without these

tip-wlan/charts/wlan-prov-service/templates/configmap.yaml
4c74356b41: does it make any sense though? whats the point of overriding them if they are defined in the global. if you override it in local to any value except the one in global it would not work

tip-wlan/charts/opensync-gw-static/templates/deployment.yaml
4c74356b41: no idea what these are for. we certainly could, probably not in this pr?

tip-wlan/charts/wlan-ssc-service/files/cqlshrc.tip-wlan
4c74356b41: its always cassandra\cassandra - so probably doesn't make a lot of sense? we don't even have a variable for that

tip-wlan/charts/opensync-gw-static/templates/deployment.yaml
brennerm: Agree with @4c74356b41. It's something we should definitely fix but not in this PR.

tip-wlan/charts/wlan-prov-service/templates/configmap.yaml
brennerm: The only use case I could think of where this is necessary is if we allow to use existing and different e.g. PSQL databases for each service. I don't think that that's a requirement for now so I'd like to prevent to make the chart even more complex for now.

tip-wlan/charts/wlan-prov-service/templates/deployment.yaml
4c74356b41: there is a very little upside to this, as this is just a big timeout before we want it to fail

CHANGELOG.md
4c74356b41: do we want a single one for both of these or not?

4c74356b41: sorry, I meant a single release for both of these PR's?

tip-wlan/templates/opensync-gw-cloud-dependencies copy.yaml
```diff
@@ -0,0 +1,17 @@
+---
```
brennerm: Is the name of this file correct?

tip-wlan/templates/opensync-mqtt-broker-dependencies.yaml
```diff
@@ -0,0 +1,11 @@
+---
```
brennerm, line 1: Shouldn't this be part of the opensync-mqtt-broker Helm chart?

tip-wlan/templates/opensync-gw-cloud-dependencies copy.yaml
brennerm: Shouldn't this be part of the opensync-gw-cloud Helm chart?

tip-wlan/charts/wlan-integrated-cloud-component-service/templates/secret.yaml
leonidlm: @4c74356b41 as we discussed a few times in our weekly calls, Max is about to work on a feature that will allow us to run the backend without TLS... So the service will work without creds.

tip-wlan/charts/wlan-prov-service/templates/configmap.yaml
leonidlm: Ok, agree with you guys on this one.

tip-wlan/charts/wlan-prov-service/templates/deployment.yaml
leonidlm: And yet, I think this should be parametrized and exposed through values for future fine tuning if someone needs it.

tip-wlan/charts/wlan-ssc-service/files/cqlshrc.tip-wlan
leonidlm: @4c74356b41 why it's always cassandra? If for example I will run this with my own instance is cassandra, I'm sure it won't use the default password. Also, the end user can specific a different user via the cassandra chart, in that case, everything here will break.

tip-wlan/charts/wlan-prov-service/templates/configmap.yaml
4c74356b41: i dont think we allow to specify different psql hostname anyway

tip-wlan/charts/wlan-ssc-service/files/cqlshrc.tip-wlan
4c74356b41: we do not support running your own instance of cassandra in the chart right now. this particular string was in this file for at least 7 months, this chart doesn't have (and never had) a variable for this user\password, it only had a variable for the user used after the initial db seeding is done. In fact I dont think the chart had the initial username\password for cassandra at any stage, so why should we change it now?

CHANGELOG.md
```diff
@@ -5,9 +5,22 @@ All notable changes to this project will be documented in this file.
...
 
-## [Unreleased]
+## [Unreleased](xxx) - 2021-02-04
 
+### Added
+
+- Replaced cassandra, postgres and kafka with upstream charts ( https://github.com/Telecominfraproject/wlan-cloud-helm/pull/49 )
+- Removed hardcoded docker secret in favor of variables ( https://github.com/Telecominfraproject/wlan-cloud-helm/pull/53 )
+
+### Replaced
```
brennerm: Replaced isn't an official type defined by keepachangelog. Would like to limit the types to [these](https://keepachangelog.com/en/1.0.0/#how).

CHANGELOG.md
```diff
@@ -5,9 +5,22 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [Unreleased](xxx) - 2021-02-04
 
+### Added
+
+- Replaced cassandra, postgres and kafka with upstream charts ( https://github.com/Telecominfraproject/wlan-cloud-helm/pull/49 )
+- Removed hardcoded docker secret in favor of variables ( https://github.com/Telecominfraproject/wlan-cloud-helm/pull/53 )
```
brennerm: It's confusing that this bullet point starts with "Removed" but is in the "Added" section.

tip-wlan/charts/wlan-integrated-cloud-component-service/templates/secret.yaml
4c74356b41: took this offline with max, agreed that this is fine for now

tip-wlan/charts/wlan-port-forwarding-gateway-service/templates/secret.yaml
4c74356b41: took this offline with max, agreed that this is fine for now

PR #59, opened 2021-02-11
adjust resource request/limit values

brennerm:
Added resource requests and limit to Kafka chart
Adjust some request values to better reflect real world usage


4c74356b41 (approved):
lgtm

PR #61, opened 2021-02-20
Wifi 1319 ssl issue

rahulsharma2020:
Updated the following:
- Added ssl.properties with additional properties which would be used to disable TLSv1.3 in Portal Service, if required (currently only needed for microK8s env).
- Moved logback.xml to under /portal/log folder. This is needed to avoid restarting respective component service after editing configMap that contains logback.xml, since with this change logback.xml is not under subPath of the mountedVolume in the Helm chart. ConfigMap updates are not passed on [if mount is in subPath](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically) 
- Updated NGINX configMap to enable logging, if needed.
- Updated Postgres image to a specific tag (in Provisioning initContainer) instead of latest.

### DIFF
--- a/tip-wlan/charts/wlan-portal-service/resources/config/ssl.properties
+++ b/tip-wlan/charts/wlan-portal-service/resources/config/ssl.properties
@@ -1,0 +1,14 @@
+truststorePass={{ .Values.global.certificatePasswords.sslTruststore }}
+truststoreFile=file:///opt/tip-wlan/certs/truststore.jks
+truststoreType=JKS
+truststoreProvider=SUN
+
+keyAlias=1
+keystorePass={{ .Values.global.certificatePasswords.sslKeystore }}
+keystoreFile=file:///opt/tip-wlan/certs/server.pkcs12
+keystoreType=pkcs12
+keystoreProvider=SunJSSE
+
+sslProtocol=TLS
+sslEnabledProtocols=TLSv1.2,TLSv1.1,TLSv1
+sslCiphers=TLS_RSA_WITH_AES_128_CBC_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256,TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA,TLS_DHE_RSA_WITH_AES_128_GCM_SHA256,TLS_DHE_RSA_WITH_AES_128_CBC_SHA256,TLS_DHE_RSA_WITH_AES_128_CBC_SHA
--- a/tip-wlan/charts/wlan-portal-service/templates/statefulset.yaml
+++ b/tip-wlan/charts/wlan-portal-service/templates/statefulset.yaml
@@ -115,7 +115,7 @@ spec:
             subPath: server.pkcs12
           - mountPath: /app/portal/log
             name: logback-config
-          {{- if .Values.disableTLSv13.enabled }}
+          {{- if not .Values.tlsv13.enabled }}
           - mountPath: /app/portal/certs
             name: ssl-config
           {{- end }}
@@ -158,7 +158,7 @@ spec:
       - name: logback-config
         configMap:
             name: {{ include "common.fullname" . }}-log-config
-      {{- if .Values.disableTLSv13.enabled }}
+      {{- if not .Values.tlsv13.enabled }}
       - name: ssl-config
         secret:
           secretName: {{ include "common.fullname" . }}-ssl-config
--- a/tip-wlan/charts/wlan-portal-service/templates/statefulset.yaml
+++ b/tip-wlan/charts/wlan-portal-service/templates/statefulset.yaml
@@ -113,9 +113,12 @@ spec:
           - mountPath: /opt/tip-wlan/certs/server.pkcs12
             name: certificates
             subPath: server.pkcs12
-          - mountPath: /app/portal/logback.xml
+          - mountPath: /app/portal/log
             name: logback-config
-            subPath: logback.xml
+          {{- if .Values.disableTLSv13.enabled }}
+          - mountPath: /app/portal/certs
+            name: ssl-config
+          {{- end }}
           - mountPath: {{ $file_store_path }}
             name: file-store-data
           {{- include "jmxPrometheus.configVolumeMount" . | nindent 10 }}
@@ -155,6 +158,11 @@ spec:
       - name: logback-config
         configMap:
             name: {{ include "common.fullname" . }}-log-config
+      {{- if .Values.disableTLSv13.enabled }}
+      - name: ssl-config
+        secret:
+          secretName: {{ include "common.fullname" . }}-ssl-config
+      {{- end }}
       {{- if not .Values.persistence.enabled }}
       - name: file-store-data
         emptyDir: {}
--- a/tip-wlan/charts/nginx-ingress-controller/templates/controller-configmap.yaml
+++ b/tip-wlan/charts/nginx-ingress-controller/templates/controller-configmap.yaml
@@ -8,3 +8,4 @@ metadata:
 data:
     external-status-address: {{ .Values.controller.config.externalStatusAddress }}
     client-max-body-size: {{ .Values.controller.config.clientMaxBodySize }}
+    error-log-level: {{ .Values.controller.config.errorLogLevel }}
--- a/tip-wlan/charts/nginx-ingress-controller/values.yaml
+++ b/tip-wlan/charts/nginx-ingress-controller/values.yaml
@@ -56,6 +56,8 @@ controller:
     ## Max message size coming from the Client
     clientMaxBodySize: "20m"
 
+    ## Error 
+    errorLogLevel: "error"
   ## It is recommended to use your own TLS certificates and keys
   defaultTLS:
     ## The base64-encoded TLS certificate for the default HTTPS server. If not specified, a pre-generated self-signed certificate is used.
--- a/tip-wlan/charts/wlan-portal-service/resources/config/logback.xml
+++ b/tip-wlan/charts/wlan-portal-service/resources/config/logback.xml
@@ -7,7 +7,7 @@
 <!-- For professional support please see                            -->
 <!--    http://www.qos.ch/shop/products/professionalSupport         -->
 <!--                                                                -->
-<configuration>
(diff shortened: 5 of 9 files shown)


AkshayJagadish-ne (approved):

brennerm:
@rahulsharma2020 Great job at finding this issue. :+1: 

I'm not quite happy with the implementation of the fix though. It's a lot of effort for something that could be configured by passing JVM options instead. Let me know what you think.

tip-wlan/charts/wlan-portal-service/resources/config/ssl.properties
```diff
@@ -0,0 +1,14 @@
...
+truststoreProvider=SUN
+
+keyAlias=1
+keystorePass={{ .Values.global.certificatePasswords.sslKeystore }}
+keystoreFile=file:///opt/tip-wlan/certs/server.pkcs12
+keystoreType=pkcs12
+keystoreProvider=SunJSSE
+
+sslProtocol=TLS
+sslEnabledProtocols=TLSv1.2,TLSv1.1,TLSv1
```
4c74356b41, line 13: why 1.0 and 1.1? those are unsecure. why no 1.3?

tip-wlan/charts/wlan-portal-service/templates/statefulset.yaml
```diff
@@ -113,9 +113,12 @@ spec:
           - mountPath: /opt/tip-wlan/certs/server.pkcs12
             name: certificates
             subPath: server.pkcs12
-          - mountPath: /app/portal/logback.xml
+          - mountPath: /app/portal/log
             name: logback-config
-            subPath: logback.xml
+          {{- if not .Values.tlsv13.enabled }}
```
4c74356b41, line 118: why is this called 1.3 is 1.3 is not actually enabled?

rahulsharma2020:
> @rahulsharma2020 Great job at finding this issue. 👍
> 
> I'm not quite happy with the implementation of the fix though. It's a lot of effort for something that could be configured by passing JVM options instead. Let me know what you think.

Thanks @brennerm. 
Yes, that's exactly (passing JVM options) is what we did. See [here](https://github.com/Telecominfraproject/wlan-cloud-services/blob/7a2772cb5d311a870f1c2c16964dccd219f1480f/portal-docker/src/main/docker/app/run.sh#L6). Currently we are passing _default_ SSL Properties as a JVM parameter (in the docker image) and overriding it (using configMap) based on the deployment. 
Are you referring to putting JVM attributes in override file(/command line during helm upgrade) or some other way? I prefer to make the deployment easy (less info in override file/command-line) and putting more complexity in the charts.
Please point out if there is a better way that I completely missed.

dmitry-toptygin-connectus-ai (approved):

tip-wlan/charts/wlan-portal-service/resources/config/ssl.properties
brennerm: There's a bug in OpenJDK with TLSv1.3. See the ticket description for further details.

rahulsharma2020: @4c74356b41: This is the solution for the bug WIFI-1319. Please see my comment [here](https://telecominfraproject.atlassian.net/browse/WIFI-1319?focusedCommentId=12837)

tip-wlan/charts/wlan-portal-service/templates/statefulset.yaml
rahulsharma2020: That's right @4c74356b41; it won't be enabled and is controlled by this flag based on environment.

brennerm:
> > @rahulsharma2020 Great job at finding this issue. +1
> > I'm not quite happy with the implementation of the fix though. It's a lot of effort for something that could be configured by passing JVM options instead. Let me know what you think.
> 
> Thanks @brennerm.
> Yes, that's exactly (passing JVM options) is what we did. See [here](https://github.com/Telecominfraproject/wlan-cloud-services/blob/7a2772cb5d311a870f1c2c16964dccd219f1480f/portal-docker/src/main/docker/app/run.sh#L6). Currently we are passing _default_ SSL Properties as a JVM parameter (in the docker image) and overriding it (using configMap) based on the deployment.
> Are you referring to putting JVM attributes in override file(/command line during helm upgrade) or some other way? I prefer to make the deployment easy (less info in override file/command-line) and putting more complexity in the charts.
> Please point out if there is a better way that I completely missed.

Yeah I've seen that PR for the portal service. What I'd prefer is to just pass a single JVM option disabling TLSv1.3 instead passing a path to a file that we have to override in the Helm chart. Would decrease complexity significantly.

rahulsharma2020:
> > > @rahulsharma2020 Great job at finding this issue. +1
> > > I'm not quite happy with the implementation of the fix though. It's a lot of effort for something that could be configured by passing JVM options instead. Let me know what you think.
> > 
> > 
> > Thanks @brennerm.
> > Yes, that's exactly (passing JVM options) is what we did. See [here](https://github.com/Telecominfraproject/wlan-cloud-services/blob/7a2772cb5d311a870f1c2c16964dccd219f1480f/portal-docker/src/main/docker/app/run.sh#L6). Currently we are passing _default_ SSL Properties as a JVM parameter (in the docker image) and overriding it (using configMap) based on the deployment.
> > Are you referring to putting JVM attributes in override file(/command line during helm upgrade) or some other way? I prefer to make the deployment easy (less info in override file/command-line) and putting more complexity in the charts.
> > Please point out if there is a better way that I completely missed.
> 
> Yeah I've seen that PR for the portal service. What I'd prefer is to just pass a single JVM option disabling TLSv1.3 instead passing a path to a file that we have to override in the Helm chart. Would decrease complexity significantly.

Ah, I see what you mean.

The way we create the SSL Connector is currently in **Base** project and shared across all the components (Portal, Provisioning, Opensync, SSC, SPC etc). Adding logic in the Base component to look for `DisableTLSv1.3` JVM attribute that would ignore particular Ciphers based on this flag is rather intrusive. We would need regression for all these components.
Moreover the 'supported-list' of Ciphers for TLSv1.2 can change; we would also need to pass them along as JVM args (which is actually a pretty big list and will look clumsy).

The current approach is making use of existing ssl.properties by adding a couple new properties (which can be edited at the deployment time - mainly referring to `ciphers` ) and only changing them in Helm chart, to minimize the impact on all-in-one-deployment as well.

Considering the timelines for this release, I would recommend staying with this approach. Let me know if I am overthinking and there is an easier, less intrusive approach?

tip-wlan/charts/wlan-portal-service/resources/config/ssl.properties
4c74356b41: okay, but why enabled 1.0 and 1.1? those are not secure

dmitry-toptygin-connectus-ai:
@brennerm : the way SSL/TLS options are configured for REST API endpoints is via per-connector properties. Not global per-JVM ones like "javax.net.ssl.*". A connector on one TCP port can use entirely different set of properties from the connector on another TCP port in the same java process. That flexibility is part of the spring framework and we are taking advantage of it. We package ssl-related properties for the connectors into ssl.properties config file. If a property is missing from the ssl.properties file, it will be taken first from the corresponding -D global system java option, and if it's missing there it will be taken from the corresponding environment variable. We could supply -DsslEnabledProtocols=... to the java command line or export sslEnabledProtocols=... before invoking java - instead of editing ssl.properties file, but that would mean that some connector properties are configured in one place while others are in a different place, which to me seems confusing.

tip-wlan/charts/wlan-portal-service/resources/config/ssl.properties
rahulsharma2020: Just giving clients ability to connect using less-secure protocols :).

Basically this issue was only seen in microK8s env and wasn't sure if we had clients using these protocols, so kept them in the list of enabledProtocol list. The system anyway chooses the most secure protocol in the list, so we should be safe. Also to add, if we don't pass this list explicitly, the system assumes support of `TLSv1.3, TLSv1.2, TLSv1.1 and TLSv1` (unless you put them in the `jdk.tls.disabledAlgorithms` list - which is not what we are doing).

Actually the `sslCiphers` list would anyway reject any TLSv1.0 or TLSv1.1 communication.

brennerm:
@rahulsharma2020 @dmitry-toptygin-connectus-ai OK I don't know how this file is being used by your service. All I know is that overwriting the whole file in the Helm chart is definitely not the right approach.

If the default ssl.properties file changes, who takes care of updating it here? Probably nobody, someone will try to run the portal service without TLSv1.3 and encounter an issue that is completely unrelated.

But from my side, go ahead.

brennerm (approved):

dmitry-toptygin-connectus-ai:
@brennerm  - I agree with you that keeping a copy of a config file fragment in helm is not good, and agree with your reasoning for it. On the other hand, I still think that keeping a single source for a particular configuration in deployment is easier to debug than to trace all possible overrides. 

@rahulsharma2020 - the change where we disallow the TLSv1.3 should be done only for the microk8s configuration, where that problem is present, it should not affect the default deployments on aws (where AFAIK this is a non-issue). 

@brennerm @leonidlm How about we create a JIRA to move away from internal config files in general and use only -D options and environment variables in helm templates? That would keep configuration consistent, and remove the need for keeping config files in helm. I propose to do it post release-1 because it is not a small endeavor. The full known list of properties used in CLoudSDK is documented in here https://telecominfraproject.atlassian.net/wiki/spaces/WIFI/pages/907411534/Configurable+properties+of+the+CloudSDK

brennerm:
+1 for the -D options or environment variables would even be better. :+1:

yongchen-cu (approved):

rahulsharma2020:
@dmitry-toptygin-connectus-ai: Yes, it's only applicable to microK8s today; controlled by flag `tlsv13.enabled`.

@brennerm: Agree with your thoughts and ownership issues pointed. Also, there is another copy of ssl.properties and httpClientConfig.json under `/tip-wlan/resources/certs/` that we should get rid of as part of the ticket that @dmitry-toptygin-connectus-ai is referring to post release-1.

4c74356b41:
well, it doesnt make any sense, how is this only applicable to microk8s? it can be deployed anywhere (by mistake or intentionally)

PR #64, opened 2021-03-01
Update statefulset.yaml

4c74356b41:
work around the imagepullbackoff issue on updates

### DIFF
--- a/tip-wlan/charts/wlan-portal-service/templates/statefulset.yaml
+++ b/tip-wlan/charts/wlan-portal-service/templates/statefulset.yaml
@@ -7,6 +7,7 @@ metadata:
   labels:
     {{- include "common.labels" . | nindent 4 }}
 spec:
+  podManagementPolicy: Parallel
   serviceName: {{ include "common.fullname" . }}
   replicas: {{ .Values.replicaCount }}
   selector:


tip-wlan/charts/wlan-portal-service/templates/statefulset.yaml
```diff
@@ -7,6 +7,7 @@ metadata:
   labels:
     {{- include "common.labels" . | nindent 4 }}
 spec:
+  podManagementPolicy: Parallel
```
brennerm, line 10: Are you sure this is going to fix anything? From [the docs](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#parallel-pod-management) this doesn't have any influence on how pods are being updated.

4c74356b41: it does fix update stuck errors in my experience

4c74356b41: https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#forced-rollback

brennerm (approved):

PR #65, opened 2021-03-05
add exporting of servo beans

brennerm:
[Link to dashboard](https://grafana.lab.wlan.tip.build/d/ryendCyMk/tip-cloud-controller-performance)

### DIFF
--- a/tip-wlan/resources/configs/jmx-prometheus-config.yml
+++ b/tip-wlan/resources/configs/jmx-prometheus-config.yml
@@ -1,13 +1,15 @@
 lowercaseOutputLabelNames: true
 lowercaseOutputName: true
-whitelistObjectNames: ["java.lang:type=OperatingSystem"]
+whitelistObjectNames:
+  - "java.lang:type=OperatingSystem"
+  - "com.netflix.servo:name=cassandra*,type=COUNTER"
+  - "com.netflix.servo:name=cassandra*,type=GAUGE"
+  - "com.netflix.servo:name=singleDataSource*,type=GAUGE"
+  - "com.netflix.servo:name=singleDataSource*,type=COUNTER"
+  - "com.netflix.servo:name=jdbc*,type=GAUGE"
+  - "com.netflix.servo:name=jdbc*,type=COUNTER"
+  - "com.netflix.servo:name=kafka*,type=GAUGE"
+  - "com.netflix.servo:name=kafka*,type=COUNTER"
+  - "com.netflix.servo:name=osgw*,type=GAUGE"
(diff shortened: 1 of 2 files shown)


leonidlm (approved):

tip-wlan/resources/configs/jmx-prometheus-config.yml
```diff
@@ -1,13 +1,15 @@
 lowercaseOutputLabelNames: true
 lowercaseOutputName: true
-whitelistObjectNames: ["java.lang:type=OperatingSystem"]
+whitelistObjectNames:
+  - "java.lang:type=OperatingSystem"
+  - "com.netflix.servo:name=cassandra*,type=COUNTER"
```
dmitry-toptygin-connectus-ai, line 5: FYI, the name matches here work according to the rules described in https://docs.oracle.com/javase/6/docs/api/javax/management/ObjectName.html

Some object names contain extra key values, and they will be matched only of ",*" is added to the patterns.
Figuring out what values to export is a balancing act - we want to see more information, but exporting all the metrics will increase the size of the dataset that prometheus polls every minute. 

Exact set of metrics to poll depends on the area under testing and on the dashboards that we want to support.

----

Examples of ObjectName patterns are:

    *:type=Foo,name=Bar to match names in any domain whose exact set of keys is type=Foo,name=Bar.
    d:type=Foo,name=Bar,* to match names in the domain d that have the keys type=Foo,name=Bar plus zero or more other keys.
    *:type=Foo,name=Bar,* to match names in any domain that has the keys type=Foo,name=Bar plus zero or more other keys.
    d:type=F?o,name=Bar will match e.g. d:type=Foo,name=Bar and d:type=Fro,name=Bar.
    d:type=F*o,name=Bar will match e.g. d:type=Fo,name=Bar and d:type=Frodo,name=Bar.
    d:type=Foo,name="B*" will match e.g. d:type=Foo,name="Bling". Wildcards are recognized even inside quotes, and like other special characters can be escaped with \.

An ObjectName can be written as a String with the following elements in order:

    The domain.
    A colon (:).
    A key property list as defined below. 

A key property list written as a String is a comma-separated list of elements. Each element is either an asterisk or a key property. A key property consists of a key, an equals (=), and the associated value.

At most one element of a key property list may be an asterisk. If the key property list contains an asterisk element, the ObjectName is a property list pattern.

Spaces have no special significance in a String representing an ObjectName. For example, the String:

 "domain: key1 = value1 , key2 = value2"

represents an ObjectName with two keys. The name of each key contains six characters, of which the first and last are spaces. The value associated with the key " key1 " also begins and ends with a space.

----

PR #67, opened 2021-03-17
make all init container images configurable

brennerm:
[WIFI-1668](https://telecominfraproject.atlassian.net/browse/WIFI-1668)

### DIFF
--- a/tip-wlan/charts/opensync-gw-cloud/values.yaml
+++ b/tip-wlan/charts/opensync-gw-cloud/values.yaml
@@ -14,17 +14,17 @@ image:
   tag: 0.0.1-SNAPSHOT
 
 waitForServicesImage:
-  registry: docker.io
+  registry: tip-docker-cache-repo.jfrog.io
   repository: opsfleet/depends-on
   tag: v1.0.0
 
 intCloudReadinessImage:
-  registry: docker.io
+  registry: tip-docker-cache-repo.jfrog.io
   repository: library/alpine
   tag: 3.13
 
 testConnectionImage:
-  registry: docker.io
+  registry: tip-docker-cache-repo.jfrog.io
   repository: library/alpine
   tag: 3.13
 
--- a/tip-wlan/charts/opensync-gw-cloud/values.yaml
+++ b/tip-wlan/charts/opensync-gw-cloud/values.yaml
@@ -16,17 +16,17 @@ image:
 waitForServicesImage:
   registry: docker.io
   repository: opsfleet/depends-on
-  tag: latest
+  tag: v1.0.0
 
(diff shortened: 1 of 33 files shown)


tip-wlan/charts/opensync-gw-cloud/values.yaml
```diff
@@ -13,6 +13,21 @@ image:
   name: opensync-gateway-cloud
   tag: 0.0.1-SNAPSHOT
 
+waitForServicesImage:
+  registry: docker.io
+  repository: opsfleet/depends-on
+  tag: latest
```
leonidlm: @brennerm maybe we should consider pinning specific versions of these images? Things can unexpectedly break if we won't do that, especially in the alpine case below.

tip-wlan/values.yaml
```diff
@@ -73,9 +73,15 @@ global:
     password: tip-read
 
   monitoring:
+    jmxExporterAgentUrl: https://repo1.maven.org/maven2/io/prometheus/jmx/jmx_prometheus_javaagent/0.14.0/jmx_prometheus_javaagent-0.14.0.jar
```
leonidlm: @brennerm are we pulling this during runtime? Maybe worth creating a Docker image for this one and use it instead of the alpine mentioned down the line in this values file?

leonidlm (approved):

tip-wlan/values.yaml
brennerm: How are we going to access the jar file from the application container in this case? :thinking:

brennerm: @leonidlm ping

leonidlm: @brennerm not 100% sure about the current implementation, but we can mount a shared ephemeral volume between two containers under one pod and just copy the jar from one filesystem to another. Saw this done multiple times in other cases.

Another option is to add this JAR to our CloudSDK Docker images, no?

tip-wlan/charts/opensync-gw-cloud/values.yaml
brennerm: Fixed

tip-wlan/values.yaml
brennerm: So we would have a Docker image that contains only the JAR file? Where's the benefit of this approach over downloading the agent from a URL on startup?

Yup, that would be another option.

brennerm: @leonidlm I'm currently waiting for Bitnami to add the Java agent JAR to their Docker image to be able to use it here. According to https://github.com/bitnami/bitnami-docker-jmx-exporter/issues/1 they will be adding it soon.

PR #69, opened 2021-03-21
automate Helm chart release process

brennerm:
Resulting release can be reviewed [here](https://github.com/Telecominfraproject/wlan-cloud-helm/releases/tag/untagged-129091d49bd265257c7b).


.github/workflows/extract-changelog-entry.py
```diff
@@ -0,0 +1,24 @@
+import sys
+
+import keepachangelog
```
leonidlm: NICE!

.github/workflows/helm-build.yml
```diff
@@ -3,28 +3,71 @@ name: Helm CI - TIP WLAN Cloud Master
...
+        with:
+          python-version: "3.8"
+      - name: create RELEASE.md
+        run: |
+          cat << EOF > RELEASE.md
+          ## Installation
+
+          \`\`\`
+          $ helm repo add tip-wlan-cloud-helm-virtual-repo https://tip.jfrog.io/artifactory/tip-wlan-cloud-helm-virtual-repo --username <USERNAME> --password <PASSWORD>
+          $ helm install tip-wlan-cloud-helm-virtual-repo/tip-wlan --version ${GITHUB_REF#refs/tags/v}
```
leonidlm: I'm not sure that I like the fact that the installation procedure is hard-coded into the build pipeline, but to be honest, I can't think of a better way? Maybe having a separate script, something like `install.sh` and then just pasting it's content here? Still not convinced it's better, though.

leonidlm (approved):
Looks good and very convenient!

.github/workflows/helm-build.yml
4c74356b41: why not put this into extract-changelog-entry.py? this way you can test this flow locally and you have an advantage of not needing this step?

I'd probably also create a container with python and keepachangelog already installed for the same reason. easier for folks to test\follow this locally

brennerm: @leonidlm Removed this section cause users need to preconfigure the environment anyway before being able to install. We can think about adding those installation instructions once the chart is actually usable in the "normal" way.

@4c74356b41 Removed this section for now. Don't see the need for a Docker image just for this purpose ATM. Don't expect anyone to execute this script locally prior to making a release.

PR #77, opened 2021-05-12
Add: values for AWS internal setup [TOOLS-136]

dunaev-dmitry:


tip-wlan/example-values/aws-internal/values.yml
```diff
@@ -0,0 +1,131 @@
+shared:
+  service:
+    srv-https-annotations: &srv-https-annotations
+      kubernetes.io/ingress.class: alb
+      alb.ingress.kubernetes.io/scheme: internet-facing
```
brennerm: Mmh, think this should be set to `internal` as well, or?

dunaev-dmitry: True, missed that somehow

dunaev-dmitry: Fixed

brennerm:
LGTM :+1:

brennerm (approved):

